PR #18, opened 2021-11-11
(ENG-822) Format menu data

thunguy:
## Description
This PR is focused on formatting menu data returned from `get_week_menu_data` by creating a wrapper function, `get_formatted_menu_data`. Notes to consider when reviewing this PR:

> 1.) What we need to know about each menuItem is:
>    - what product_code does it relate to on the thistle-web side (represented by `itemCode` in the expected return value e.g. LV1 or DM3 or RSK)
>    - what top-level recipe is it related to (represented by `recipeId`, e.g. "cmVjaXBlOjE3ODAxMA==" is the id for "Thistle Pad Thai with Sesame Baked Tofu" recipe)
>    - what standalone item is associated with the top-level recipe (represented by `standaloneRecipeId` e.g. "Thistle Pad Thai with Sesame Baked Tofu" recipe has a standalone item with an id of "cmVjaXBlOjE3NDQxMg==" to represent "Chili Tamarind Sauce")
>    - any non-standalone recipe will have a `None` value for `standaloneRecipeId`

> 2.) We only need the `recipeItems` so that we can identify any 'standalone' recipe. Once we've identified it, we just want its id, which should be returned as `standaloneRecipeId`.

> 3.) There will be a max of **one** `standaloneRecipeId` per menu item within `menuItems` list, if any.

## Test Plan
Open python interactive within galley-sdk: `$ python`
`>>> from pprint import pprint` (optional: to apply pretty print format to returned data)
`>>> from galley.queries import *` 

### 1. All valid menu names passed in:
`>>> pprint(get_formatted_menu_data(["2021-10-04 1_2_3", "2021-10-25 1_2_3"]))`

> **Expectation:** A return value **similar** to:
> ```python3
> [
>  {'name': '2021-10-04 1_2_3',
>   'date': '2021-10-04',
>   'id': 'bWVudTo2MzQ2NTY=',
>   'location': 'Vacaville',
>   'menuItems': [{'itemCode': None,
>                  'recipeId': 'cmVjaXBlOjE3NDc5MA==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 's2',
>                  'recipeId': 'cmVjaXBlOjE3Mjk3NQ==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 'dm3',
>                  'recipeId': 'cmVjaXBlOjE3ODAxOQ==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 'dv3',
>                  'recipeId': 'cmVjaXBlOjE3ODAxMA==',
>                  'standaloneRecipeId': 'cmVjaXBlOjE3NDQxMg=='},
>                 {'itemCode': None,
>                  'recipeId': 'cmVjaXBlOjE3NDc5Nw==',
>                  'standaloneRecipeId': None},
>                 { ... # total len of menuItems should be 25 with 3 containing a 'standaloneRecipeId' }]},
>  {'name': '2021-10-25 1_2_3',
>   'date': '2021-10-25',
>   'id': 'bWVudTo3NjM2MjA=',
>   'location': 'Vacaville',
>   'menuItems': [{'itemCode': 'b1',
>                  'recipeId': 'cmVjaXBlOjE4MDcxMA==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 'lv1',
>                  'recipeId': 'cmVjaXBlOjE4MjEzOA==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 'lm1',
>                  'recipeId': 'cmVjaXBlOjE4MjE0Ng==',
>                  'standaloneRecipeId': None},
>                 {'itemCode': 'dv1',
>                  'recipeId': 'cmVjaXBlOjE3OTk1Mw==',
>                  'standaloneRecipeId': 'cmVjaXBlOjE3NDI2MQ=='},
>                 {'itemCode': None,
>                  'recipeId': 'cmVjaXBlOjE4MjI0NQ==',
>                  'standaloneRecipeId': 'cmVjaXBlOjE3NDI2MQ=='},
>                 {'itemCode': 'dm1',
>                  'recipeId': 'cmVjaXBlOjE3OTk2MA==',
>                  'standaloneRecipeId': 'cmVjaXBlOjE3NDI2MQ=='},
>                 {'itemCode': 's1',
>                  'recipeId': 'cmVjaXBlOjE4MDk2Mg==',
>                  'standaloneRecipeId': None},
>                 { ... # total len of menuItems should be 42 with 4 containing a 'standaloneRecipeId' }]}
> ]
>```

### 3. Some valid menu names passed in:
`>>> pprint(get_formatted_menu_data(["2021-10-04 1_2_3", "2021-10-25 1_2_3", "invalid-menu-name 1_2_3"]))`

> **Expectation:** Same return value as above

### 4. Invalid or no menu names passed in:
`>>> pprint(get_formatted_menu_data(["invalid-menu-name 4_5_6"]))`
> **Expectation:** A return value of `[]`

`>>> pprint(get_formatted_menu_data([]))`
> **Expectation:** A return value of `[]`

### 5. Run unittests
Exit python interactive and stay within the galley-sdk directory to run tests:
`(galley-env) $ python -m unittest tests/test_queries.py`
**[OPTIONAL SANITY CHECK]** `(galley-env) $ python -m unittest tests/test_*.py`

## Versioning
Updated to v0.11.0

### DIFF
--- a/galley/queries.py
+++ b/galley/queries.py
@@ -81,3 +81,45 @@ def get_week_menu_data(names: list) -> Optional[List[Dict]]:
     raw_data = make_request_to_galley(op=query.__to_graphql__(auto_select_depth=3), variables={'name': names})
     return validate_response_data(raw_data, 'menus')
 
+
+def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
+    menus = get_week_menu_data(names)
+    formatted_menus = []
+
+    if not menus:
+        return None
+
+    for menu in menus:
+        formatted_menu = ({
+            'name': menu.get('name'),
+            'id': menu.get('id'),
+            'date': menu.get('date'),
+            'location': menu['location'].get('name'),
+            'menuItems': []
+        }) # type: Dict
+
+        menu_items = menu.get('menuItems', [])
+
+        for menu_item in menu_items:
+            recipe_items = menu_item.get('recipe', {}).get('recipeItems', [])
+
+            formatted_menu['menuItems'].append({
+                'itemCode': next((cv.get('name') for cv in menu_item['categoryValues']), None),
+                'recipeId': menu_item.get('recipeId'),
+                'standaloneRecipeId': get_standalone(recipe_items)
+            })
+
+        formatted_menus.append(formatted_menu)
+    return formatted_menus
+
+
+# Returns the subRecipeId if any 'standalone' item exists within recipeItems, else returns None
+# It is assumed that there is a max of ONE standalone item within the list of recipeItems, if any.
+def get_standalone(recipe_items):
+    for recipe_item in recipe_items:
+        preparations = recipe_item.get('preparations', [])
+        is_standalone = any(prep['name'] == 'standalone' for prep in preparations)
+
+        if is_standalone:
+            return recipe_item.get('subRecipeId')
+    return None
--- a/galley/queries.py
+++ b/galley/queries.py
@@ -9,8 +9,6 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-
-
 class Viewer(Type):
     recipes = Field(Recipe, args=(ArgDict({'where': FilterInput})))
     recipe = Field(Recipe, args={'id': str})
@@ -60,8 +58,8 @@ def recipes_data_query(recipe_ids: List[str]) -> Optional[Operation]:
     query.viewer.recipe.recipeItems.__fields__('ingredient', 'subRecipe', 'preparations')
     query.viewer.recipe.recipeItems.ingredient.__fields__('externalName', 'categoryValues')
     query.viewer.recipe.recipeItems.ingredient.categoryValues.__fields__('name')
-
     return query
+
 
 def get_raw_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
     query = recipes_data_query(recipe_ids=recipe_ids)
@@ -99,7 +97,6 @@ def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
         }) # type: Dict
 
         menu_items = menu.get('menuItems', [])
-
         for menu_item in menu_items:
             recipe_items = menu_item.get('recipe', {}).get('recipeItems', [])
 
--- a/galley/queries.py
+++ b/galley/queries.py
@@ -82,6 +82,6 @@ def menu_data_query(names: List[str]) -> Optional[Operation]:
 
 
 def get_raw_menu_data(names: List[str]) -> Optional[List[Dict]]:
-    query = menu_data_query(names=names)
+    query = menu_data_query(names=names) # type: Operation
     raw_data = make_request_to_galley(op=query.__to_graphql__(auto_select_depth=3), variables={'name': names})
     return validate_response_data(raw_data, 'menus')
--- a/galley/queries.py
+++ b/galley/queries.py
@@ -67,7 +67,9 @@ def get_raw_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
     return validate_response_data(raw_data, 'recipes')
 
 
-def get_week_menu_data(names: list) -> Optional[List[Dict]]:
+# MENU QUERIES
+
+def menu_data_query(names: List[str]) -> Optional[Operation]:
     query = Operation(Query)
     query.viewer.menus(where=FilterInput(name=names)).__fields__(
         'id', 'name', 'date', 'location', 'menuItems'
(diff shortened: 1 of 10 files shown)


tests/test_queries.py
```diff
@@ -476,14 +496,65 @@ def test_get_week_menu_data_validation_failure(self, mock_retrieval_method):
...
+                    'standaloneRecipeId': None
+                }, {
+                    'itemCode': 'lm2',
+                    'recipeId': 'RECIPE3GHI',
+                    'standaloneRecipeId': 'SUBRECIPEID321'
+                }]
+            })
 
-class TestQueryRecipeIngredients(TestCase):
+        mock_retrieval_method.side_effect = [
```
drewdas, line 527: why is this a side effect? as the method is being mocked here it should be a return_value, usually side effects are used for raising exceptions

tests/test_queries.py
```diff
@@ -476,14 +496,65 @@ def test_get_week_menu_data_validation_failure(self, mock_retrieval_method):
...
+        self.assertEqual(result3, [formatted_menu('21-11-28 456')])
+
+        # one invalid menu name
+        result4 = get_formatted_menu_data(['21-12-05 456'])
+        self.assertEqual(result4, None)
+
+    @mock.patch('galley.queries.make_request_to_galley')
+    def test_get_formatted_menu_data_null(self, mock_retrieval_method):
+        mock_retrieval_method.return_value = None
+        result = get_formatted_menu_data([])
```
drewdas, line 553: if the return value is none, wouldnt None be passed to the method get_formatted_menu_data ?

galley/queries.py
```diff
@@ -129,3 +129,46 @@ def get_formatted_recipe_ingredients(recipe_id) -> Optional[Dict]:
         return {'ingredients': main_ingredients, 'standalone_ingredients': standalone_ingredients}
     else:
         return None
+
+
+def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
+    menus = get_week_menu_data(names)
```
drewdas, line 135: we are already formatting the data in get_week_menu_data by calling the method validate_response_data, if we have agreed that this is how the menu data should be formatted, should we consider just formatting the data before its returned in get_week_menu_data itself?

galley/queries.py
```diff
@@ -129,3 +129,46 @@ def get_formatted_recipe_ingredients(recipe_id) -> Optional[Dict]:
...
+    formatted_menus = []
+
+    if not menus:
+        return None
+
+    for menu in menus:
+        formatted_menu = ({
+            'name': menu.get('name'),
```
drewdas, line 143: do we want getting the keys here to fail silently? for example if 
`a = {}` `a.get('name')` would just return `None` hiding the issue that the data is missing the key. if we are expecting the data to contain certain keys we should have it throw KeyError so that we can handle that exception. here's an example.

```
try:
    name = menu['name']
except KeyError:
    logger.error('menu data for date x is missing key "name"')
    name = ""
```

this way the method would still return blank values if key does not exist but if we were investigating why menu data is missing a certain piece of data we can actually check the logs.

if you think a key is crucial such then the method should just raise an exception and not return a blank value.

drewdas (changes requested):

tests/test_queries.py
thunguy: I used side effect here to return a different value every call.

thunguy: https://docs.python.org/3/library/unittest.mock.html#:~:text=side_effect%3A%20A%20function,the%20return%20value.

tests/test_queries.py
thunguy: We can't explicitly pass a `None` type into `get_formatted_menu data()`.

thunguy: If the data is None, then it gets implicitly handled by `validate_response_data()` https://github.com/Thistle/galley-sdk/blob/5f5e8b178595b9002f36772924c14898ea4d070f/galley/common.py#:~:text=def%20validate_response_data(data,return%20None

thunguy: Would it be more appropriate to change the name of the test function?

galley/queries.py
thunguy: I believe the reason why we don't want to modify the original query function, and instead create a wrapper function, is because we want to account for the different data we'll need returned, say for thistle-web vs. okra for example.

thunguy: It's also in the notes  for the ticket that it should be split out into its own wrapper function. 

Explanation being: 

> "Instead of updating get_week_menu_data()directly, we most likely want to create a wrapper function to handle the formatting. The wrapper function is what would be called by a consuming service like thistle-web, and would itself call get_week_menu_data(). This is a pattern we have started with recipe ingredient data, via get_formatted_recipe_ingredients().  Doing this will make it easier to move the formatting code over to thistle-web, if we decide later that we'd rather have the formatting happen there." 

https://linear.app/thistle/issue/ENG-822/format-menu-data#:~:text=Instead%20of%20updating,formatting%20happen%20there.

galley/queries.py
thunguy: This makes sense in case it fails on our end. For context, it's required that all the fields within `formatted_menu` (minus `date`) are non-nullable on the Galley side.

thunguy: I'm not necessarily sure if we should have them return as empty strings, but I could be wrong! @mollykmcglone What do you think?

tests/test_queries.py
thunguy: https://docs.python.org/3/library/unittest.mock.html#:~:text=If%20side_effect%20is%20an%20iterable%20then%20each%20call%20to%20the%20mock%20will%20return%20the%20next%20value%20from%20the%20iterable.

galley/queries.py
mollykmcglone: I think this is a case where we want to support the happy path for now and come back later to add handling for missing data. We will need to do the same at the recipe level as well.

tests/test_queries.py
drewdas: yeah I think I got confused by the name 😄

galley/queries.py
drewdas: sounds good to me, we are checking validity in a method at a higher level so I'm fine with that

galley/queries.py
drewdas: ah I see, thank you

drewdas (approved):
LGTM

PR #19, opened 2021-11-11
(ENG-823) refactor raw and formatted recipe queries to combine data

smunro-thistle:
## Description
This change combines recipe ingredient, nutrition, and basic metadata into a single, batched query. It also moves methods that format recipe data into a new file to distinguish between actual graphql queries and data manipulation.

Question: I'm interested in thoughts on how we should organizing mocked data. I pulled it out into separate files because we need similar mocked data across two test suites (test_queries and test_formatted_queries). They're python files because some of the data is hierarchical and I wanted to be able to import the lower level data into the higher level files.

## Test Plan
in interactive python, run:
```
import pprint; import galley

result = get_formatted_recipes_data(["cmVjaXBlOjE3OTk1Mw==", "cmVjaXBlOjE3Mjg4Mg==", "FAKE!!!"])
pp = pprint.PrettyPrinter()
pp.pprint(result)

```

You can compare these results (for the real recipes) to the recipes on galley. Note that all packaging and standalone items should be excluded from the ingredients list.

[https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3OTk1Mw==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3OTk1Mw==)
[https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3Mjg4Mg==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3Mjg4Mg==)


setup.py
```diff
@@ -4,7 +4,7 @@
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.8.0',
+    version='0.9.0',
```
thunguy, line 7: We can bump this to v0.10.0 since we're currently on v0.9.0

galley/formatted_queries.py
```diff
@@ -0,0 +1,71 @@
...
+    
+    return ingredients
+
+def get_formatted_recipes_data(recipe_ids: list) -> Optional[List[Dict]]:
+    recipes_data = get_raw_recipes_data(recipe_ids=recipe_ids) or []
+    formatted_recipes = []
+    
+    for recipe in recipes_data:
+        formatted_recipe = {}
+        formatted_recipe['id'] = recipe.get('id')
```
talkallday, line 44: NB: I have a strong stylistic preference for 
```
formatted_recipe = {
    'id': recipe.get('id'),
    'externalName': recipe.get('externalName')
    ...
}
````

the repeated `formatted_recipe` is a lot of unnecessary characters and makes it harder for me to read

File: galley/queries.py [hunk 3]
```diff
@@ -8,8 +8,7 @@
 
 logger = logging.getLogger(__name__)
 
-FOOD_PACKAGING = 'food pkg'
```
talkallday, line 11: hm wouldn't it be safer and better for mypy to keep the variable values instead of returning to directly calling the strings each time we need them?

galley/formatted_queries.py
```diff
@@ -0,0 +1,71 @@
...
+            if preparations:
+                is_standalone = any(prep.get('name') == 'standalone' for prep in preparations)
+
+            if not is_standalone:                    
+                for ingredient in item_ingredients:
+                    if ingredient not in ingredients:
+                        ingredients.append(ingredient)
+    
+    return ingredients
+
```
talkallday, line 37: should be two spaces between function definitions

galley/formatted_queries.py
```diff
@@ -0,0 +1,71 @@
+from typing import Dict, Optional, List
+
+from galley.queries import get_raw_recipes_data
+
+
+def ingredients_from_recipe_items(recipe_items: list) -> Optional[List]:
```
talkallday, line 6: the type annotation here should be `List[Dict]`. i'm actually not sure how you're getting away with using `list` here at all 🤔

galley/queries.py
```diff
@@ -53,11 +52,19 @@ def get_recipe_data() -> Optional[List[Dict]]:
     return validate_response_data(raw_data, 'recipes')
 
 
-def get_recipe_nutrition_data(recipe_ids: list) -> Optional[List[Dict]]:
+def recipes_data_query(recipe_ids: list) -> Optional[Operation]:
```
talkallday, line 55: type annotation should be `recipe_ids: List[int]`

galley/queries.py
```diff
@@ -53,11 +52,19 @@ def get_recipe_data() -> Optional[List[Dict]]:
...
-        'id', 'externalName', 'notes', 'description', 'categoryValues', 'reconciledNutritionals'
+        'id', 'externalName', 'notes', 'description', 'categoryValues', 'reconciledNutritionals', 'recipeItems'
     )
+    query.viewer.recipe.recipeItems.__fields__('ingredient', 'subRecipe', 'preparations')
+    query.viewer.recipe.recipeItems.ingredient.__fields__('externalName', 'categoryValues')
+    query.viewer.recipe.recipeItems.ingredient.categoryValues.__fields__('name')
+
+    return query
+
+def get_raw_recipes_data(recipe_ids: list) -> Optional[List[Dict]]:
```
talkallday, line 66: type annotation should be recipe_ids: List[int]

galley/formatted_queries.py
mollykmcglone: Sure, will update

galley/queries.py
mollykmcglone: I'm not sure about Sarah's thinking in removing these vars, (maybe because they are each only being used in one place at this time?) I will add them back for now, I think we'll eventually be using them more than in a few contexts.

galley/queries.py
mollykmcglone: I will update to be more specific, but it will be List[str] as recipe_ids in Galley are strings

galley/queries.py
mollykmcglone: I will update to be more specific, but it will be List[str] as recipe_ids in Galley are strings

galley/queries.py
mollykmcglone: added these back in the new file, formatted_queries.py

galley/formatted_queries.py
```diff
@@ -0,0 +1,70 @@
...
+
+            if not is_standalone:                    
+                for ingredient in item_ingredients:
+                    if ingredient not in ingredients:
+                        ingredients.append(ingredient)
+    
+    return ingredients
+
+
+def get_formatted_recipes_data(recipe_ids: list) -> Optional[List[Dict]]:
```
talkallday, line 41: type annotation should be `List[int]` here

talkallday: actually no for galley recipe ids `List[str]` right?

mollykmcglone: yeah they are strings, just updated

galley/formatted_queries.py
```diff
@@ -0,0 +1,70 @@
...
+                is_standalone = any(prep.get('name') == STANDALONE for prep in preparations)
+
+            if not is_standalone:                    
+                for ingredient in item_ingredients:
+                    if ingredient not in ingredients:
+                        ingredients.append(ingredient)
+    
+    return ingredients
+
+
```
talkallday, line 40: now that the pagination decorator is in `main`, maybe rebase and add it here?

setup.py
```diff
@@ -4,7 +4,7 @@
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.9.0',
+    version='0.10.0',
```
talkallday, line 7: i think now this should be 0.11.0

galley/formatted_queries.py
talkallday: remember to add parentheses!

```
@paginate_results()
def get_formatted_recipes_data(...)
```

mollykmcglone: done!

talkallday (approved):

PR #20, opened 2021-11-15
(ENG-857) add pagination decorator

talkallday:
## Description
adding a pagination decorator so we can abstract from concerns about response size limitations from galley

## Test Plan
automated tests

## Versioning
incrementing to 0.10.0

### DIFF
--- a/galley/pagination.py
+++ b/galley/pagination.py
@@ -1,0 +1,24 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def paginate_results(func, page_size: int = 25):
+    """
+    can decorate any function with an initial list input
+    and collect its output in pages of page_size
+    """
+
+    def wrapper(*args):
+        input_list = args[0]
+        result = []
+        if type(input_list) is not list:
+            logger.error(f'argument {input_list} is not a list')
+            return result
+        for partition in range((len(input_list)//page_size) + 1):
+            result.extend(
+                func(input_list[partition * page_size: (partition + 1) * page_size])
+            )
+        return result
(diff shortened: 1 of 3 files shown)


thunguy (changes requested):
Automated tests passed, but an issue occurred when I ran manual tests on `get_recipe_ingredients()` and `get_recipe_nutrition_data()` with a list of 26 unique `recipe_ids`: 

First attempt (failed): I added `@paginate_results` (with no explicit page_size) and I was returned the decorator function.

Second attempt (success):  I added `@paginate_results(page_size=2)`and I was returned the expected result.

I assume that we'd want the same results to return regardless of an explicit page_size input for an arbitrary amount of recipe_ids passed in > 25.

talkallday:
> Automated tests passed, but an issue occurred when I ran manual tests on `get_recipe_ingredients()` and `get_recipe_nutrition_data()` with a list of 26 unique `recipe_ids`:
> 
> First attempt (failed): I added `@paginate_results` (with no explicit page_size) and I was returned the decorator function.
> 
> Second attempt (success): I added `@paginate_results(page_size=2)`and I was returned the expected result.
> 
> I assume that we'd want the same results to return regardless of an explicit page_size input for an arbitrary amount of recipe_ids passed in > 25.

we were able to resolve -- to get default 25 need to add parentheses to decorator use then behavior is as expected 
```
@paginate_results()
def some_function(a_list):
...
```

thunguy (approved):
LGTM! Automated and manual tests passed with the expected results returned.

PR #21, opened 2021-11-16
(ENG-823) Bug Fix - corrects viewer.recipe to viewer.recipes for batch query

smunro-thistle:
## Description
This updates the query to correctly batch the request for multiple recipe ids. We should now be correctly filtering packaging, whereas previously packaging ingredients were being included in the ingredients list. It also updates the test to correct the expected query.
## Test Plan
In interactive python, run:

`import pprint; import galley`

`pprint(get_formatted_recipes_data(["cmVjaXBlOjE3OTk1Mw==", "cmVjaXBlOjE3Mjg4Mg==", "FAKE!!!"]))`

You can compare these results to the recipes on galley. Note that all packaging and standalone items should be excluded from the ingredients list.

https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3OTk1Mw==
https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3Mjg4Mg==

## Versioning
Updated to 0.11.1

### DIFF
--- a/galley/queries.py
+++ b/galley/queries.py
@@ -55,9 +55,9 @@ def recipes_data_query(recipe_ids: List[str]) -> Optional[Operation]:
     query.viewer.recipes(where=FilterInput(id=recipe_ids)).__fields__(
         'id', 'externalName', 'notes', 'description', 'categoryValues', 'reconciledNutritionals', 'recipeItems'
     )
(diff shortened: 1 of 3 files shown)


dip-thistle (approved):
Tested it locally. Works as expected! Approved

PR #22, opened 2021-11-17
Drewdas/eng 821 galley menu data retrieval method refactor

drewdas:
## Description

https://linear.app/thistle/issue/ENG-821/set-up-caching-for-menu-data

- Changes menu query to take list of dates instead of names
- added filtering based on location and menu type for menu data
- added menu category enum to hold id values
- expanded category and category values fields to return ids
- adjusted existing tests to reflect the above

## Test Plan
- check automated tests
- to test method in python console,`from galley.queries import get_raw_menu_data` then pass a list of dates, can also be filtered based on menu type and location
- to test filtering on menu type duplicate a menu and then change tag menu type in galley to development.

## Versioning

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -1,6 +1,6 @@
 from typing import Dict, Optional, List
 
-from galley.queries import get_raw_recipes_data, get_raw_menu_data
+from galley.queries import MenuCategoryEnum, get_raw_recipes_data, get_raw_menu_data
 from galley.pagination import paginate_results
 
 FOOD_PACKAGING = 'food pkg'
@@ -103,6 +103,12 @@ def get_formatted_menu_data(dates: List[str],
             'location': menu['location'].get('name'),
             'menuItems': []
         }) # type: Dict
+
+        categoryValues = menu['categoryValues']
+        for categoryValue in categoryValues:
+            if (categoryValue['category']['id'] ==
+                    MenuCategoryEnum.MENU_TYPE.value):
+                formatted_menu['categoryMenuType'] = categoryValue['name']
 
         menu_items = menu.get('menuItems', [])
         for menu_item in menu_items:
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -85,7 +85,6 @@ def get_formatted_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
     return formatted_recipes
 
 
-@paginate_results()
 def get_formatted_menu_data(dates: List[str],
                             location_name: str="Vacaville",
                             menu_type: str="production"
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -86,8 +86,8 @@ def get_formatted_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
 
 
 @paginate_results()
-def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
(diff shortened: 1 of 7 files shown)


galley/formatted_queries.py
```diff
@@ -86,8 +86,8 @@ def get_formatted_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
 
 
 @paginate_results()
-def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
-    menus = get_raw_menu_data(names)
+def get_formatted_menu_data(dates: List[str]) -> Optional[List[Dict]]:
```
mollykmcglone, line 89: hey @drewdas am wondering about location- I think it might make more sense to accept a location param at this (higher) level that defaults to Vacaville, and then get_raw_menu_data() can just have location as a required param. That way we can build location in to the calling function in thistle-web now. Was there a reason you didn't set up a location param here?

mollykmcglone:
I have a question/suggestion about how we're handling location, but otherwise this looks great to me! I pulled the branch down locally and tested, saw the expected results.

galley/formatted_queries.py
drewdas: yeah good catch, I think it makes sense to access params here at this level too, I missed it because it got recently added

drewdas: @mollykmcglone implemented the change

galley/formatted_queries.py
```diff
@@ -86,8 +86,11 @@ def get_formatted_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
 
 
 @paginate_results()
-def get_formatted_menu_data(names: list) -> Optional[List[Dict]]:
-    menus = get_raw_menu_data(names)
+def get_formatted_menu_data(dates: List[str],
+                            location_name: str="Vacaville",
```
mollykmcglone, line 90: Sending a non-Vacaville location_name arg is not working for me (Vacaville gets set, regardless). I did a little debugging and I think the problem is in `paginate_results()` decorator. It looks like that's written to work only with one arg that is a List. So I think we'll either need to drop the pagination here (which I think would be ok for now) or refactor that decorator.

drewdas: yeah thats not properly implemented, I removed it and added tests to assert what args get_formatted_menu_data is called with so that when we add it back there are regression tests

mollykmcglone: Great, thanks for adding those tests!

mollykmcglone (approved):
Code and tests look good! I pulled the latest down locally and was able to retrieve menus and filter by location or type as expected.

galley/queries.py
```diff
@@ -3,16 +3,26 @@
...
 logger = logging.getLogger(__name__)
 
+FOOD_PACKAGING = 'food pkg'
+STANDALONE = 'standalone'
+
+class MenuCategoryEnum(Enum):
+    """
+    Enum for categories, for item type menu <category name>: <category id>
+    """
+    MENU_TYPE = 'Y2F0ZWdvcnk6MjQ2NQ=='
```
dip-thistle, line 20: Will it always be this unique ID?

drewdas: Good question, They have consistent values across menus if you look at the category section in galley they have the same ID in the url. As of whether this unique ID will change I am not sure. @mollykmcglone can we reach out to galley to check if the category ids can possibly change at some point?

PR #23, opened 2021-11-18
Eng 869 menu upsert

patrick-thistle:
## Description
This branch adds the functionality to make menu/menuItem upsert requests specifically to set volume counts for existing menuItems present.

## Test Plan
Unittests are present for all new functionality.

## Versioning
Version incremented from 0.11.0 -> 0.12.0

### DIFF
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -47,11 +47,11 @@ def build_unit_input(item):
 def build_menu_item_inputs(items):
     menu_item_inputs = []
     for item in items:
-        unitInput = build_unit_input(item)
+        unit_input = build_unit_input(item)
         menuItemInput = MenuItemInput(
             id=item['id'],
             volume=item['volume'],
-            unit=unitInput
+            unit=unit_input
         )
         menu_item_inputs.append(menuItemInput)
     return menu_item_inputs
@@ -62,17 +62,17 @@ def build_menu_inputs(menus):
         if 'menu_id' not in menu.keys() or menu['menu_id'] is None:
             raise ValueError("menu_id is required for successful requests")
 
-        menuInput = MenuInput(id=menu['menu_id'])
-        menuInput.menuItems = build_menu_item_inputs(menu['menu_items'])
-        menu_inputs.append(menuInput)
+        menu_input = MenuInput(id=menu['menu_id'])
+        menu_input.menuItems = build_menu_item_inputs(menu['menu_items'])
+        menu_inputs.append(menu_input)
 
     return menu_inputs
 
 def build_upsert_mutation_query(args):
     mutation = Operation(Mutation)
-    bulkInput = BulkMenusInput()
-    bulkInput.menus = build_menu_inputs(args)
-    mutation.bulkUpsertMenus(input=bulkInput)
+    bulk_input = BulkMenusInput()
+    bulk_input.menus = build_menu_inputs(args)
+    mutation.bulkUpsertMenus(input=bulk_input)
     return mutation
 
 # args = [
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -35,9 +35,12 @@ def add_recipe_instruction(recipe_id, instruction, position=None):
     return validate_response_data(raw_data, 'createRecipeInstruction', 'recipeInstruction')
 
 # MENU MUTATIONS
-
+UNIT_NAME_WHITELIST = ['each']
 def build_unit_input(item):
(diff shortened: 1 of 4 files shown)


galley/types.py
```diff
@@ -151,6 +160,24 @@ class Menu(Type):
...
+class MenuItemInput(Input):
+    id = str
+    recipeId = str
+    volume = float
+    unit = UnitInput
+
+class MenuInput(Input):
+    id = str
+    recipeId = str
+    menuItems = Field(list_of(MenuItemInput))
```
smunro-thistle, line 175: I'm trying to get a better understanding of sgqlc. When do we need to use Field() and when don't we? Just if it's another graphql Type/Input?

tests/test_mutations.py
```diff
@@ -72,3 +77,180 @@ def test_add_recipe_instruction_validation_failure(self, mock_mr):
...
+    @mock.patch('galley.mutations.make_request_to_galley')
+    def test_upsert_menu_data_returns_expected_response_data(self, mock_mr):
+        mock_mr.return_value = self.response_data
+        payload = [{
+            "menu_id": "bWVudTo5ODIzOTI=",
+            "menu_items": [
+                {
+                    "id": "bWVudUl0ZW06MzM4MTQyMjQ=",
+                    "volume": 20,
+                    "unit_name": "each"
```
smunro-thistle, line 215: Maybe a test where the unit_name is not "each"? To make sure your overwrite works

tests/test_mutations.py
```diff
@@ -72,3 +77,180 @@ def test_add_recipe_instruction_validation_failure(self, mock_mr):
...
+                }
+            ]
+        }]
+        result = upsert_menu_data(payload)
+        # import pdb; pdb.set_trace()
```
smunro-thistle, line 220: leftover debugging?

smunro-thistle (changes requested):
A couple NB questions, and then one small change recommended for a test.

galley/types.py
patrick-thistle: I'll try my best to interpret my understanding from the SGQLC typing documentation here - https://sgqlc.readthedocs.io/en/latest/sgqlc.types.html#sgqlc.types.Field

in the instance above "Field" is a container required for the nesting of differnent types within an interface.  In this case Field() was wrapped around list_of due to an error returned during manual testing that the attribute was not present on the model.

The documentation seems to imply that an implicit Field wrapper is placed around the base types (in this case str).  We could technically write the interface as `id = Field(str)`

tests/test_mutations.py
patrick-thistle: The override was removed as a part of the MVP.  This will now default to a unit_type of each regardless and the unit_name is not a necessary parameter.

galley/mutations.py
```diff
@@ -31,3 +34,53 @@ def add_recipe_instruction(recipe_id, instruction, position=None):
...
+    return menu_item_inputs
+
+def build_menu_inputs(menus):
+    menu_inputs = []
+    for menu in menus:
+        if 'menu_id' not in menu.keys() or menu['menu_id'] is None:
+            raise ValueError("menu_id is required for successful requests")
+
+        menuInput = MenuInput(id=menu['menu_id'])
+        menuInput.menuItems = build_menu_item_inputs(menu['menu_items'])
```
smunro-thistle, line 61: NB: In the queries we've been doing camel case to match galley. Not sure if we want to do that here too?

tests/test_mutations.py
patrick-thistle: Scratch that.  I have added the functionality back into the PR and a new spec is in place!

tests/test_mutations.py
patrick-thistle: Rrrrremoved!

galley/mutations.py
patrick-thistle: I've gone ahead and updated this to match pythonic styles for variable names as this is explicitly a .py file.

Seeing as how this is an sdk to be consumed by a python app and is written in python it makes the most sense to me to match that styleguide when not explicitly interacting with building graphql payloads.

smunro-thistle: That sounds good! I think we should have a forum conversation about where in the flow to switch from camel case (galley's graphql format) to snake case (python standard)

smunro-thistle (approved):
Looks good! Tested locally and all works as expected on galley's staging env

PR #25, opened 2021-11-24
Eng 882

patrick-thistle:
## Description
Update lookups for Category and Preparations to take place by ID not by name.  The primary location where a name lookup for preparations were taking place was with the is_standalone() and is_packaging check for RecipeItem.

This branch also updates any queries that look for preparations to return the preparation's ids in addition to name, same goes for ingredient categories.

## Test Plan
I'm not 100% of all the places that needed to be updated as a part of this work.  I have updated all the existing tests which check STANDALONE and is_packaging to ensure that they now pass.  There is no net new functionality to go along with these lookups, so theoretically if the tests pass everything is still working as expected.

## Versioning
[X] Please update the project version in setup.py


drewdas (approved):
LGTM!

mollykmcglone (approved):
Code looks great, I also pulled down this branch and manually tested recipe weight and ingredients list, all good!

PR #27, opened 2021-11-30
(ENG-870) add menu item id to menu data retrieval and formatting functions

mollykmcglone:
## Description
This adds another attribute, menu item id, to the data that gets retrieved from Galley and returned by get_formatted_menu_data. We will be using menu item id within thistle-web to send menu item projections to Galley. Tests and mock data are updated to account for the new attribute.

## Test Plan
Open python interactive within galley-sdk: $ python
`from pprint import pprint` (optional: to apply pretty print format to returned data)

Request formatted menu data for a couple weeks:
`pprint(get_formatted_menu_data(['2022-01-03', '2021-12-27']))`

Validate that for each menu item included in the returned menus, an id is present (can additionally use the Galley playground to validate that the value of the ids matches expectations).
## Versioning
Please update the project version in setup.py -> Done!

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -168,6 +168,7 @@ def get_formatted_menu_data(dates: List[str],
             recipe_items = menu_item.get('recipe', {}).get('recipeItems', [])
 
             formatted_menu['menuItems'].append({
+                'id': menu_item.get('id'),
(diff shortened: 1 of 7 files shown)


dip-thistle (approved):
Tested locally! Approved

PR #29, opened 2021-12-01
Refactor recipes data retrieval to include standalone fields

drewdas:
…ight and standalone recipe information

## Description
## Test Plan
## Versioning


tests/mock_responses/mock_recipe_tree_components.py
```diff
@@ -421,3 +421,578 @@
...
+                    "name": "each"
+                },
+                "value": 1
+            }
+        ],
+        "id": "l7n5ecjjdyd",
+              "ingredient": None,
+              "recipeItem": {
+                  "subRecipeId": "cmVjaXBlOjE3MDM5Mw==",
+            "subRecipe": {
```
thunguy, line 440: Is `subRecipe` a part of `recipeItem` here?  It seems as if the indentation may be off.

galley/formatted_queries.py
```diff
@@ -75,21 +94,49 @@ def get_recipe_category_tags(recipe_category_values: List[Dict]) -> Optional[Dic
...
+
+    standalone_subrecipe = standalone_recipe_item.get('subRecipe', {})
+
+    return {
+        'weight': round(total_weight, 2),
+        'standaloneRecipeId': standalone_subrecipe.get('id'),
+        'standaloneRecipeName': get_external_name(standalone_subrecipe),
+        'standaloneNutrition': standalone_subrecipe.get('reconciledNutritionals'),
+        'standaloneIngredients': standalone_subrecipe.get('allIngredients'),
+        'standaloneWeight': standalone_recipe_item.get('quantity'),
```
mollykmcglone, line 137: We need the weight here in grams (and don't need the standaloneWeightUnit field, below). One way to do that could be by setting standalone_weight at line 118 by adding a line like `standalone_weight = recipe_item.mass()`. Then at line 137 you could do `'standaloneWeight': round(standalone_weight, 2)`

thunguy (approved):
LGTM! Tested locally and got expected results as laid out in the Test Plan. Also ran automated tests and all unittests passed.

galley/formatted_queries.py
```diff
@@ -75,21 +94,49 @@ def get_recipe_category_tags(recipe_category_values: List[Dict]) -> Optional[Dic
...
+        logger.error("More than one standalone recipe items found for recipe"
+                     f"tree component id {recipe_tree_component.get('id')}")
+
+    standalone_subrecipe = standalone_recipe_item.get('subRecipe', {})
+
+    return {
+        'weight': round(total_weight, 2),
+        'standaloneRecipeId': standalone_subrecipe.get('id'),
+        'standaloneRecipeName': get_external_name(standalone_subrecipe),
+        'standaloneNutrition': standalone_subrecipe.get('reconciledNutritionals'),
```
mollykmcglone, line 135: Just did some manual testing and this handling will result in returning the nutritionals for the batched recipe, as opposed to the single serving size. This is because we're using the subrecipe here as opposed to the recipe_item from within recipe_tree_components. 

I think we need to set this value above, for example at line 118 you could add `standalone_nutrition = recipe_item_dict.get('reconciledNutritionals')`. Then here you can just return that value, i.e. `'standaloneNutrition': standalone_nutrition`. I know this is a subtle distinction and am glad to talk through it/look at an example recipe in Galley together if that would be helpful.

tests/mock_responses/mock_recipe_tree_components.py
drewdas: yeah subrecipe should be indented here, I'll go ahead and fix it

mollykmcglone (approved):
Looks good, I tested manually and saw the expected values for the top-level and standalone recipes. Nice work @drewdas !

PR #30, opened 2021-12-14
ENG-901 Use RecipeConnection for pagination of recipes

smunro-thistle:
## Description
In order to implement more flexible pagination, we are switching from using the plain recipe model to Galley's recipeCollection model. The get_raw_recipes_data method has been updated to use the recipe_connection_query, which allows for pagination. get_raw_recipes_data aggregates the pages so that it returns recipes in the same format as its previous version.

The max page size it 25 items.

Tests and mock data have been updated.

## Test Plan
On line 94 of queries.py, update page_size to another size (like 2)

In the python console:

```
from galley.queries import *
import pprint; import galley

recipe_ids = ["cmVjaXBlOjE2NDgzNQ==", "cmVjaXBlOjE2NTE3NA==", "cmVjaXBlOjE2NTE3NQ==", "cmVjaXBlOjE2NTE3Ng==", "cmVjaXBlOjE2NTE3OQ=="]

recipe_data = get_raw_recipes_data(recipe_ids=recipe_ids)
```

And check that recipe_data returns the expected results, in the expected format. These results should look the same as the previous, unpaginated results.


mollykmcglone (approved):
Looks good! I pulled the branch down and tested as described and with `get_formatted_recipes_data(recipe_ids)`, and saw the expected results.

PR #31, opened 2021-12-21
(maint) get standaloneNutrition from subrecipe instead of recipeItem

mollykmcglone:
## Description
We're changing how we handle standalone components within Galley and as a result need to access nutrition data for standalone components differently. This updates format_recipe_tree_components_data() to get the value of standaloneNutrition from the standalone subrecipe data layer, as opposed to the standalone recipeItem data layer. Tests and fixtures are updated to account for the change.

## Test Plan
To test this change, you will first want to identify a standalone recipe to work with, i.e. [Peanut Coconut Sauce](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3MDM5NA==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D), and a parent recipe the standalone is used in, i.e. [Balinese Gado Gado Salad with Sesame Baked Tofu](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE2NzEwOQ==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D)

Then use the interactive shell to pull the formatted recipe details for the parent recipe.
`from pprint import pprint`
`pprint(get_formatted_recipes_data(['cmVjaXBlOjE2NzEwOQ==']))`

On the main branch, the standaloneNutrition returned from that request will match up with the nutrition as calculated for the usage amount of the standalone component in the parent recipe, no matter what the nutrition serving size is set to on the standalone recipe nutrition tab (so 2.5oz in the case of the Gado Gado Salad). On this branch, the standaloneNutrition will match up with what the nutrition serving size is set to on the standalone nutrition tab (1 oz in the case of the Peanut Coconut Sauce).

## Versioning
Please update the project version in setup.py: DONE


tests/mock_responses/mock_recipe_tree_components.py
```diff
@@ -501,77 +501,77 @@
             "subRecipe": {
                 "id": "cmVjaXBlOjE3MDM5Mw==",
                 "name": "Balinese Gado Gado Salad BASE",
-                        "externalName": None,
```
mollykmcglone, line 504: Noticed the indentation was off in here, these are not functional changes

tests/mock_responses/mock_recipe_tree_components.py
```diff
@@ -867,77 +867,77 @@
...
-                            "vitaminDMcg": 0,
-                            "vitaminDPercentRDI": None,
-                            "vitaminEMg": 426.0233792000001,
-                            "vitaminEPercentRDI": 28.402,
-                            "vitaminKMcg": 44.5345236,
-                            "vitaminKPercentRDI": 0.371,
-                            "zincMg": 118.49965336831083,
-                            "zincPercentRDI": 10.773
-                        }
+                    "addedSugarG": 0,
```
mollykmcglone, line 881: I replaced this nutritional data for the Peanut Coconut Sauce, which was scoped to a full batch of sauce, with data scoped to a single serving.  That  matches up with how we intend to set up these recipes moving forward.

dip-thistle (approved):

PR #33, opened 2021-12-28
(maint) itemCode referenced before assignment fix

talkallday:
## Description
in cases where product code is not assigned to a menu item an `UnboundLocalError` is being thrown

## Test Plan
in shell access a menu with a menu item that does not have a product code assigned to it. currently there is one in staging for 10-04-2021

```
from galley.formatted_queries import get_formatted_menu_data
dates = ['2021-10-04']
get_formatted_menu_data(dates=dates)
```

on main, this will generate an `UnboundLocalError` that will go away with this fix

## Versioning


mollykmcglone (approved):
Tested locally and worked as expected!

PR #34, opened 2021-12-28
Eng 943 add recipe meal slug

patrick-thistle:
## Description
This branch pulls the base meal slug and includes it in the formatted responses for menu.

## Test Plan
You can check the meal slugs are returned with the formatted queries like so.
```
import galley
from galley.formatted_queries import get_formatted_menu_data
get_formatted_menu_data(['2022-01-03'])
```

## Versioning
[X] Please update the project version in setup.py


smunro-thistle (approved):
Looks good! Tested locally and mealSlug returns as expected.

PR #35, opened 2021-12-29
(maint) add hasStandalone and baseMeal fields to formatted recipe data

mollykmcglone:
## Description
This adds two new fields to the formatted data returned for each recipe. hasStandalone is a boolean field representing whether the recipe includes a standalone component. This new field will allow us to validate on fields specific to the standalone recipe when expected. baseMeal is a categoryValue (tag) being set on each recipe in Galley that we forgot to include in the formatted data previously. Tests are updated to cover the expected behaviors.

## Test Plan
To test this, retrieve recipe details for a couple recipes (one w/standalone and one without) and ensure that the `hasStandalone` and `baseMeal` fields are returned with the expected values for each recipe.

This example uses the Balinese Gado Gado Salad (with standalone) and Burmese Yellow Curry(no standalone):

`from pprint import pprint`
`pprint(get_formatted_recipes_data(['cmVjaXBlOjE2NzEwOQ==', 'cmVjaXBlOjE3NjQ5Mg==']))`

## Versioning


dip-thistle (approved):
Let's get this merged!

PR #36, opened 2022-01-06
(ENG-932) add handling for dynamic standalone serving sizes

mollykmcglone:
## Description
This updates get_formatted_recipes_data() and its underlying functions to return a few additional fields about standalone components: standaloneSuggestedServing, standaloneServingSizeWeight, and standaloneServings. 

This new handling allows us to provide more than one serving of a standalone component within a recipe, and display accurate information to customers on labels and in our website as to how many servings of the standalone component are provided. Previously our implementation assumed that we would only ever provide one serving of standalone components. 

Tests are added/updated to account for the expected behaviors.

## Test Plan
To test this, retrieve recipe details for a few different cases: a recipe without any standalone component, a recipe with multiple servings of a standalone component,  and a recipe with one serving of a standalone component. Ensure that the standaloneSuggestedServing, standaloneServingSizeWeight, and standaloneServings fields are returned with the expected values for each recipe.

This example uses the Balinese Gado Gado Salad (has multiple servings of a standalone component), Crunchy Thai Peanut Bowl (has one serving of a standalone component), and Burmese Yellow Curry(no standalone component):

`from pprint import pprint`
`pprint(get_formatted_recipes_data(['cmVjaXBlOjE2NzEwOQ==', 'cmVjaXBlOjE5MDkwNA==', 'cmVjaXBlOjE3NjQ5Mg==']))`

## Versioning

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -22,7 +22,34 @@ def get_external_name(data_dict):
             return data_dict['name']
     return None
 
+def calculate_servings(usage_quantity: Optional[float], nutritionals_quantity: Optional[float]) -> Optional[float]:
+    """
+    Given a usage quantity (how much of a given component is included in a recipe)
+    and a nutritionals_quantity (the size of one serving of the component),
+    returns a number representing how many servings of the component
+    are included in the recipe.
+    """
+    if usage_quantity is not None and nutritionals_quantity is not None:
+        return usage_quantity/nutritionals_quantity
+    else:
+        return None
 
+def calculate_serving_size_weight(weight: Optional[float], number_of_servings: Optional[float]) -> Optional[float]:
+    """
+    Given a weight (representing the total weight of the component included in a recipe)
+    and a number of servings (how many servings of the component are included),
+    returns a number representing the weight of just one serving of the component.
+    """
+    if weight is not None and number_of_servings is not None:
+        return weight/number_of_servings
+    else:
+        return None
+
+def format_suggested_serving(quantity: Optional[float], unit: Optional[str]) -> Optional[str]:
+    if quantity is not None and unit is not None:
+        return "{} {}".format(quantity, unit)
+    else:
+        return None
 
 class RecipeItem:
     def __init__(self, preparations: List[Dict[str, str]], ingredient=None,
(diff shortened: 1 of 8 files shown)


galley/formatted_queries.py
```diff
@@ -139,21 +190,33 @@ def format_recipe_tree_components_data(recipe_tree_components: List[Dict]) -> Di
...
             standalone_data['standaloneNutrition'] = standalone_subrecipe.get('reconciledNutritionals')
             standalone_data['standaloneIngredients'] = standalone_subrecipe.get('allIngredients')
-            standalone_data['standaloneWeight'] = round(standalone_recipe_item_weight, 2) if standalone_recipe_item_weight else None
+            standalone_data['standaloneWeight'] = round(standalone_recipe_item_weight) if standalone_recipe_item_weight else None
+            standalone_data['standaloneSuggestedServing'] = "{} {}".format(standalone_nutritionals_quantity, standalone_nutritionals_unit)
+            standalone_data['standaloneServingSizeWeight'] = round(standalone_serving_size_weight) if standalone_serving_size_weight else None
+            standalone_data['standaloneServings'] = standalone_servings if standalone_servings else None
 
     return {
-        'weight': round(total_weight, 2),
```
mollykmcglone, line 156: This is an unrelated change- we've found that we just need the whole number of grams for weight so updated the rounding here, and related tests.

galley/formatted_queries.py
```diff
@@ -44,12 +71,40 @@ def is_packaging(self):
...
+            unit = self.subrecipe.get('nutritionalsUnit')
+            if unit is not None:
+                result = unit.get('name')
+        return result
+
+    def standalone_nutritionals_quantity(self):
+        if self.is_standalone and self.subrecipe is not None:
+            return self.subrecipe.get('nutritionalsQuantity')
+        else:
+            return None
```
dip-thistle, line 92: NB: Can we combine these two methods into one? The reason is it has the same condition for both so this feels like repetition of a logic.

galley/formatted_queries.py
```diff
@@ -44,12 +71,40 @@ def is_packaging(self):
...
+
+    def standalone_usage_quantity(self):
+        """
+        Returns the recipe item's usage quantity (how much of a given component is included in a recipe)
+        based on the type of unit specified by the nutritonals_unit (i.e. "oz"). Returns None if there is
+        not a quantity availabe for the specified unit.
```
dip-thistle, line 98: NB: Typo on spelling `available`

tests/test_formatted_queries.py
```diff
@@ -33,27 +35,146 @@ def test_ingredients_from_recipes_empty(self):
...
+    def test_calculate_servings_returns_None_if_usage_quantity_missing(self):
+        expected_result = None
+        result = calculate_servings(None, 1.0)
+        self.assertEqual(result, expected_result)
+        
+    def test_calculate_servings_returns_None_if_nutritionals_quantity_missing(self):
+        expected_result = None
+        result = calculate_servings(2.5, None)
+        self.assertEqual(result, expected_result)
+
```
dip-thistle, line 53: NB: Two spaces between class definition.

dip-thistle (commented):
I will be testing the PR locally now. Some comments based on code review.

galley/formatted_queries.py
mollykmcglone: I'd prefer separate methods, but agree the conditional logic is the same- how about I split that off to make this DRYer? Something like `if has_standalone_subrecipe:`

mollykmcglone: Just pushed that change, let me know if think that's sufficient or if you'd prefer a different approach!

dip-thistle (approved):
Let's get this merged!

PR #37, opened 2022-01-18
(ENG-990) Pull photos for recipe and return lifestylePhotoUrl

smunro-thistle:
## Description
Updated galley-sdk to pull a recipe's 'media' attribute (a list of RecipeMedia objects) in recipe_connection_query. Updated get_formatted_recipes_data to search the RecipeMedia objects for an object with the caption 'lifestyle'. It then returns the sourceUrl for this RecipeMedia object as the formatted recipe's lifestylePhotoUrl.

## Test Plan
1. Make sure all tests are passing
2. Test raw query locally in python console (make sure both images show up in the media attribute):
note: The recipe ID provided is a [fake recipe](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE5OTgzOA==) I've modified to have two images, one that has the caption 'lifestyle'. This will be overwritten if this PR isn't reviewed before the overnight job replaces staging data.
```
from galley.queries import *
get_raw_recipes_data(recipe_ids=["cmVjaXBlOjE5OTgzOA=="])
```
3. Test formatted data locally (make sure the correct image is returned as the lifestylePhotoUrl:
```
get_formatted_recipes_data(recipe_ids=["cmVjaXBlOjE5OTgzOA=="])
```

## Versioning

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -115,6 +115,7 @@ class FormattedRecipe:
         self.externalName = get_external_name(recipe_data)
         self.notes = recipe_data.get('notes')
         self.description = recipe_data.get('description')
+        self.lifestylePhotoUrl = get_lifestyle_photo(recipe_data.get('media', []))
         self.nutrition = recipe_data.get('reconciledNutritionals', {})
         self.recipe_category_values = recipe_data.get('categoryValues', [])
         self.recipe_tags = get_recipe_category_tags(self.recipe_category_values)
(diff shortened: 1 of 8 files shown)


galley/formatted_queries.py
```diff
@@ -281,6 +283,13 @@ def get_meal_slug(menu_item: Dict) -> Optional[str]:
     return None
 
 
+def get_lifestyle_photo_url(media: List) -> Optional[str]:
+    for photo in media:
+        if photo.get('caption', '') == RecipeMediaEnum.LIFESTYLE_CAPTION.value and photo.get('sourceUrl', None):
+            return photo.get('sourceUrl')
+    return None
```
drewdas, line 290: NB: somewhere down the line we'll probably need a default image if this url is None, should we add an environment variable here such as 404_MEAL_LIFESTYLE_PHOTO_URL? might be too early to worry about it

drewdas (approved):
LGTM

PR #41, opened 2022-02-01
(ENG-998) add name to menu item recipe

talkallday:
## Description
in 0.24.0, we pulled only `externalName` for menu item recipes, but many recipes have only a `name`, so adding that to the query

## Test Plan
Compared to `main`, reviewer can see in shell that this adds `recipeName` value for many menu items with this command
```
>>> from galley.formatted_queries import get_formatted_menu_data
>>> import pprint
>>> pp = pprint.PrettyPrinter()
>>> pp.pprint(get_formatted_menu_data(['2022-01-31']))

## Versioning


patrick-thistle (approved):
approved.

PR #42, opened 2022-02-18
(ENG-1077) handle non-sellable menu items

mollykmcglone:
## Description
This adds support for filtering non-sellable menu items from the menu data retrieved from Galley. An optional parameter, onlySellableMenuItems, is added to get_formatted_menu_data() so that callers can exclude non-sellable menu items from the returned menus, if desired. Tests and mock data are updated accordingly.

## Test Plan
To test, open the python shell and pull a menu that has at least one non-sellable item on it (I used the [2022-03-07 1_2_3](https://staging-app.galleysolutions.com/menus/bWVudToxNDU1MzI1) menu for testing, which currently has one non-sellable item). Compare the menu_items returned when the new optional parameter is not set vs set, you should see the non-sellable menu item(s) be filtered out as expected.

For example:
`menu = get_formatted_menu_data(['2022-03-07'])`
`len(menu[0]['menuItems'])` -> 44
`menu = get_formatted_menu_data(['2022-03-07'], onlySellableMenuItems=True)`
`len(menu[0]['menuItems'])` -> 43

## Versioning

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -148,6 +148,7 @@ class FormattedRecipe:
         self.externalName = get_external_name(recipe_data)
         self.notes = recipe_data.get('notes')
         self.description = recipe_data.get('description')
+        self.isSellable = recipe_data.get('isDish')
         self.lifestylePhotoUrl = get_lifestyle_photo_url(recipe_data.get('media', []))
         self.nutrition = recipe_data.get('reconciledNutritionals', {})
         self.recipe_category_values = recipe_data.get('categoryValues', [])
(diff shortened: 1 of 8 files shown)


galley/formatted_queries.py
```diff
@@ -399,25 +401,27 @@ def get_formatted_menu_data(dates: List[str],
         menu_items = menu.get('menuItems', [])
         for menu_item in menu_items:
             formatted_recipe = FormattedRecipe(recipe_data=menu_item.get('recipe', {}))
+            skip_non_sellable = onlySellableMenuItems and not formatted_recipe.isSellable
```
talkallday, line 404: would you be opposed to just using continue here? like

``` 
if onlySellableMenuItems and not formatted_recipe.isSellable:
    continue
    ```

talkallday: then you wouldn't need the `skip_non_sellable` variable at all nor the indent below

mollykmcglone: yeah I'm cool with that, will update!

PR #43, opened 2022-03-11
(ENG-1101) Add gross weight to recipes

smunro-thistle:
## Description
We need to differentiate between net weight (weight excluding standalone and packaging) and gross weight (weight including standalone and packaging). This PR renames weight to netWeight, adds the calculation for gross weight, and updates our mocks and tests accordingly.

## Test Plan
1. Look through test updates and mocks and make sure that calculated net/gross weights are as expected
2. Test get_formatted_recipes_data on a recipe on galley staging with packaging and standalone component and confirm that the grossWeight returned matches the total weight on the dashboard. The example I've compared against is: https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE4MjEzOA==

Not that grossWeight is returned in grams, so you'll have to convert to lbs to compare

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -35,8 +35,8 @@ def calculate_servings(
     nutritionals_quantity: Optional[float]
 ) -> Optional[float]:
     """
-    Given a usage quantity (how much of a given component 
-    is included in a recipe) and a nutritionals_quantity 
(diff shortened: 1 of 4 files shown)


mollykmcglone (approved):
I reviewed the code and tests/mocks and tested this locally, am seeing the expected behavior!

PR #45, opened 2022-04-06
(ENG-987) Retrieve ops menu data for Okra

dip-thistle:
## Description
Since Okra's data is going to be constructed based off of Galley menu data, and Okra's data needs are significantly different than thistle-web's in terms of recipe data, this PR adds two separate functions `get_ops_menu_data` & `get_formatted_ops_menu_data` to query Galley's API and return the fields needed to construct our Okra entities / screens. 

Review playground query and data requirements for Okra [here](https://paper.dropbox.com/doc/Okra-Required-Fields-For-Galley--BY9hk8z~hb~WpIgI9w1i1p2UAg-A1ejqyyRmnfYS41FWdtMQ#:h2=Mapping-1:1) to compare the results returned from the query.

## Test Plan
```
(galley-env) python
>>> import galley; from galley.formatted_queries import *; from galley.queries import *; from pprint import pprint
>>> pprint(get_formatted_ops_menu_data(["2022-03-28", "2022-03-31"]))
```

## Run Automated Tests
```
(galley-env) python -m unittest tests/test_*
```

## Versioning


tests/test_queries.py
```diff
@@ -608,3 +619,120 @@ def test_get_raw_recipes_multiple_pages(self, mock_retrieval_method):
         result = get_raw_recipes_data(['1', '2', '3'])
         self.assertEqual(mock_retrieval_method.call_count, 2)
         self.assertEqual(result, expected_recipe_data)
+
+
+class TestOpsMenuDataQuery(TestCase):
+    def setUp(self) -> None:
+        self.expected_query = '''query {
```
patrick-thistle, line 626: We really should find a better pattern than this... 😬

patrick-thistle: Maybe we can just call this class good enough for it's own file and create a test_ops_queries.py file?

tests/test_formatted_queries.py
```diff
@@ -725,3 +716,88 @@ def test_get_formatted_menu_data_args_defaults(self, mock_grmd):
...
+            'mealContainer': 'ts32',
+            'mealCode': 'dv3',
+            'recipeTreeComponents': mock_recipeTreeComponents,
+            'totalCount': 549
+        }]
+    }
+    return formatted_ops_menu
+
+
+class TestGetFormattedOpsMenuData(TestCase):
```
patrick-thistle, line 769: Can we break this class out into it's own test file as well..?

galley/formatted_queries.py
```diff
@@ -367,6 +358,20 @@ def get_menu_photo_url(media: List) -> Optional[str]:
...
+        if (category_value['category']['id'] == MenuItemCategoryEnum.PRODUCT_CODE.value):
+            return category_value['name']
+    return ''
+
+
+def get_category_menu_type(menu_category_values) -> str:
+    for category_value in menu_category_values:
+        if (category_value['category']['id'] == MenuCategoryEnum.MENU_TYPE.value):
+            return category_value['name']
+    return ''
```
patrick-thistle, line 372: I know that both of these methods are tested in passing via the big bulk JSON comparison, but it would be really helpful to just have a couple of smaller direct tests on new methods so we're not strictly reliant on the big fragile monster comparisons.

tests/test_queries.py
thunguy: Done!

tests/test_formatted_queries.py
thunguy: Done!

galley/formatted_queries.py
thunguy: Done!

patrick-thistle (approved):
Tested and everything appears to match up to spec.

LGTM!

PR #46, opened 2022-04-11
(ENG-873) retrieve highlight tags from Galley

mollykmcglone:
## Description
This adds a new field, "highlight_tags", to the response provided by get_formatted_recipes_data(). The highlight_tags field is an array of string values representing what has been assigned to the recipe for the highlight_1 and highlight_2 categories. If no highlight tags have been applied to the recipe, the highlight_tags field is expected to be an empty array.

## Test Plan
To manually test that the new "highlight_tags" field is behaving as expected, I updated recipe tags as necessary (as there aren't many tags already applied to recipes) and then went into the python shell and pulled recipe data for those recipes: 
`python`

`pprint(get_formatted_recipes_data(['cmVjaXBlOjE2NzEwOQ==']))` -> [Gado Gado Salad with Tofu](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE2NzEwOQ==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D) (2 tags)

`pprint(get_formatted_recipes_data(['cmVjaXBlOjE5MDE2Mw==']))` -> [Cherry Mint Pistachio Salad](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE5MDE2Mw==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D) (1 tag)

`pprint(get_formatted_recipes_data(['cmVjaXBlOjE4OTMzMw==']))` -> [Jollof Rice Bowl with Pea Protein Crumbles](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE4OTMzMw==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D) (no tags) 

## Versioning
- [x ] Please update the project version in setup.py

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -217,6 +217,8 @@ def get_recipe_category_tags(
         RecipeCategoryTagTypeEnum.PROTEIN_ADDON_TAG.value: 'proteinAddOn',
         RecipeCategoryTagTypeEnum.BASE_MEAL_SLUG_TAG.value: 'baseMealSlug',
         RecipeCategoryTagTypeEnum.BASE_MEAL_TAG.value: 'baseMeal',
+        RecipeCategoryTagTypeEnum.HIGHLIGHT_ONE_TAG.value: 'highlightOne',
+        RecipeCategoryTagTypeEnum.HIGHLIGHT_TWO_TAG.value: 'highlightTwo',
     }
 
     for recipe_category_value in recipe_category_values:
@@ -226,7 +228,23 @@ def get_recipe_category_tags(
 
         if label and recipe_category_value_name:
             recipe_tags.setdefault(label, recipe_category_value_name)
-    return recipe_tags
+
+    return format_highlight_tags(recipe_tags)
+
+def format_highlight_tags(
+    recipe_tags: Dict
+) -> Optional[Dict]:
+    highlight_tag_values = []
+    new_recipe_tags = dict(recipe_tags)
+
+    for key, value in new_recipe_tags.items():
+        if key in ('highlightOne', 'highlightTwo'):
+            highlight_tag_values.append(value)
+
+    new_recipe_tags.pop('highlightOne', None)
+    new_recipe_tags.pop('highlightTwo', None)
+    new_recipe_tags.setdefault('highlightTags', highlight_tag_values)
+    return new_recipe_tags
 
 
 def format_recipe_tree_components_data(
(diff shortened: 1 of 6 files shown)


galley/formatted_queries.py
```diff
@@ -226,7 +228,23 @@ def get_recipe_category_tags(
...
             recipe_tags.setdefault(label, recipe_category_value_name)
-    return recipe_tags
+
+    return format_highlight_tags(recipe_tags)
+
+def format_highlight_tags(
+    recipe_tags: Dict
+) -> Optional[Dict]:
+    highlight_tag_values = []
+    new_recipe_tags = dict(recipe_tags)
```
patrick-thistle, line 238: Were you running into issues where you had to recast recipe_tags as a dictionary?

galley/formatted_queries.py
```diff
@@ -226,7 +228,23 @@ def get_recipe_category_tags(
 
         if label and recipe_category_value_name:
             recipe_tags.setdefault(label, recipe_category_value_name)
-    return recipe_tags
+
+    return format_highlight_tags(recipe_tags)
```
patrick-thistle, line 232: At first pass it was confusing to me that you start with all these tags and then just returned what appeared to be only highlight tags... maybe if there was the potential for additional tag formatters down the line a more verbose approach is more clear?  In the least it would save a trip to another method in order to discover that more than just highlight tags are returned. 🤷 

```
recipe_tags = format_highlight_tags(recipe_tags)
(hypothetical)
recipe_tags = some_future_tag_modifier(recipe_tags)
return recipe_tags
```

galley/formatted_queries.py
```diff
@@ -226,7 +228,23 @@ def get_recipe_category_tags(
...
+
+def format_highlight_tags(
```
patrick-thistle, line 234: This method appears to get tested in passing for examples that contain 2 highlight tags ('new', 'spicy')

It would be useful to create some stand alone method tests for to ensure both one and none cases continue as expected.

galley/formatted_queries.py
mollykmcglone: This is likely not necessary. I have been playing around with taking a more functional approach to writing code, and copying instead of directly updating the dict fell in line with that. Do you have a preference/concern about the copy?

galley/formatted_queries.py
mollykmcglone: Yeah, good call. I went back and forth with how to incorporate the highlight tag formatting, I think what you've laid out is less confusing. Will update!

galley/formatted_queries.py
mollykmcglone: Good catch! I will add those

galley/formatted_queries.py
mollykmcglone: FU on this: I went ahead and removed the copy

patrick-thistle (approved):
Thank you for the additional specs!

LGTM!

PR #47, opened 2022-04-15
(ENG-1122) retrieve noNutritionOnWebsite tag

mollykmcglone:
**NOTE: This PR should not be reviewed until after Saturday, 4/16. The new tag being referenced here was created and applied to recipes via Galley's production instance on Friday 4/15 and will not be available for testing on their staging instance until their regular nightly job copies over the production data.**

## Description
This retrieves and handles a new tag that is applied to recipes in Galley when we cannot display their nutritional information or ingredient lists on thistle.co. When the tag has been applied to a recipe, the value of the displayNutritionOnWebsite attribute returned by get_formatted_recipe_details() is False. When the tag is not applied to a recipe, the value of displayNutritionOnWebsite defaults to True. Tests are updated to account for the new behavior.

## Test Plan
To manually test that the new "displayNutritionOnWebsite" field is behaving as expected, I went into the python shell and pulled recipe data for a regular recipe that should have a (default) True value for this field ([Gado Gado Salad with Tofu](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE2NzEwOQ==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D)) and then for a soup recipe that should have a False value ([Minestrone soup](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE4MzgyNQ==/?userLocationId=bG9jYXRpb246MTkyOA%3D%3D))

`python`

`pprint(get_formatted_recipes_data(['cmVjaXBlOjE2NzEwOQ==']))` -> Gado Gado Salad with Tofu
`pprint(get_formatted_recipes_data(['cmVjaXBlOjE4MzgyNQ==']))` -> Minestrone soup

## Versioning


setup.py
```diff
@@ -4,7 +4,7 @@
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.29.0',
+    version='0.30.0',
```
thunguy, line 7: This will need to be updated as we're currently on v0.30.0 at the moment.

thunguy (approved):
LGTM! All manual tests checked out as described, and all automated tests passed.

PR #51, opened 2022-05-11
(VEG-34) retrieve more recipe details at menu level

mollykmcglone:
## Description
This updates `get_formatted_menu_data()` so that it returns the underlying recipe's `allergens` and `highlightTags` for each menuItem in the formatted menu. This also fixes a bug that was causing the `recipeMenuPhotoUrl` attribute on menuItems to always be an empty value, and removes the unused standaloneRecipeId attribute on menuItems. Tests and mock data are updated to confirm the expected behaviors.

## Test Plan
To manually test that the new fields are being returned as expected, go into the python shell and pull formatted menu data. Confirm that each menuItem has the new fields `highlightTags`, `hasAllergen`, and `allergens`, and that the `recipeMenuPhotoUrl` field is not always returning None. The exact value of the fields will depend on the state of the recipe in Galley- not many recipes in Galley have menu photos or highlightTags yet. The [24 Carrot Gold Muffin](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3Mjk3NQ==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D) is a useful test recipe as it has an allergen, a menu photo, and a highlight tag. It's present on the 2022-05-09 menu as itemCode s2.

`python`
`pprint(get_formatted_menu_data(['2022-05-09'], onlySellableMenuItems=True))`

## Versioning


tests/test_formatted_queries.py
```diff
@@ -769,6 +777,7 @@ def test_get_formatted_menu_data_successful_for_one_valid_menu(self, mock_retrie
 
     @mock.patch('galley.queries.make_request_to_galley')
     def test_get_formatted_menu_data_successful_for_multiple_valid_menus(self, mock_retrieval_method):
+        self.maxDiff = None
```
talkallday, line 780: NB: is it preferable to return this to the default after debugging? or does it matter at all?

tests/mock_responses/mock_menu_data.py
```diff
@@ -78,6 +78,23 @@ def mock_menu(date, location_name="Vacaville", menu_type="production"):
...
+                            'sourceUrl': 'https://cdn.filestackcontent.com/Recipe_1_Plating',
+                            'mediaId': 'bWVkaWE6TEOwNQ==',
+                            'storageKey': 'Thistle/Media/TnLOVsiTG61I5W1VFwqw_.jpg'
+                        },
+                        {
+                            'altText': 'Recipe_1_Menu.jpg',
+                            'caption': 'menu',
+                            'sourceUrl': 'https://cdn.filestackcontent.com/Recipe_1_Menu',
+                            'mediaId': 'bWVkaWE6OTEwNA==',
+                            'storageKey': f'Thistle/Media/1uTFWcWhTIGBpybJ1axc_.jpg'
```
ErikBrown, line 94: NB: Unnecessary `f` here and on other storage keys below.

ErikBrown (approved):
Confirmed that `highlightTags`, `hasAllergen`, and `allergens` present on menuItems, no longer returning `standaloneRecipeId`, and there is at least one record that has a `recipeMenuPhotoUrl` without `None`.

talkallday (approved):
all worked well for me! i just wasn't see all menu photo urls as null in main so i wasn't sure where the fix was there, but the results are as expected and the code makes sense!

tests/mock_responses/mock_menu_data.py
mollykmcglone: thanks, will remove!

tests/test_formatted_queries.py
mollykmcglone: good catch, I have typically removed this after debugging (and will do now)!

PR #52, opened 2022-05-12
(maint) fix type error in format_bin_weight()

mollykmcglone:
## Description
This tweaks the logic within format_bin_weight() to be a bit simpler and resolves a Type Error that was being thrown in local environments. The expected behavior and return values of format_bin_weight() are unchanged and look to be adequately covered by tests, so no test changes are included here.

## Test Plan
To confirm that this has no impact on the overall binWeight behavior, you could go into the python shell and run the `get_formatted_ops_menu_data()` function (which returns a `binWeight` attribute for recipeComponents):

`python`
`from galley.formatted_ops_queries import *`
`pprint(get_formatted_ops_menu_data(["2022-05-09", "2022-05-12"]))`

## Versioning
- [x ] Please update the project version in setup.py

### DIFF
--- a/galley/formatted_ops_queries.py
+++ b/galley/formatted_ops_queries.py
@@ -131,16 +131,18 @@ def format_bin_weight(category_values: List) -> Dict:
     returns 'lb'.
     """
     tags = set([RecipeCTagEnum.BIN_WEIGHT_TAG.value, IngredientCTagEnum.BIN_WEIGHT_TAG.value])
-    weight = {
-        'value': DEFAULT_BIN_WEIGHT_VALUE,
-        'unit': DEFAULT_BIN_WEIGHT_UNIT
-    }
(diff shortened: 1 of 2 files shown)


galley/formatted_ops_queries.py
```diff
@@ -131,16 +131,17 @@ def format_bin_weight(category_values: List) -> Dict:
...
-    }
+    value = DEFAULT_BIN_WEIGHT_VALUE
+
     if category_values:
         for cv in category_values:
             if cv.get('category', {}).get('id') in tags:
                 value = float(cv['name'])
-                return weight | {'value': value} # type: ignore
-    return weight
+
```
thunguy, line 140: Would we want to `break` once the bin value weight has been set?

mollykmcglone: ah, yeah- I see now that's what the previous return statement was aiming for, will update!

PR #56, opened 2022-06-12
FRU-83 Create Bulk Mutation to for adding "Core Recipe" Preparation tags

ugonwigwethistle:
## Description
This features adds a bulk mutation function to the adds a "Core Recipe" Preparation tag to all BASE Recipe Items provided. The script requires as input, a set of recipe item IDs provided by the Culinary OPS, locates all recipe items in Galley, and for each the base recipe (Usage Recipe) and appends a "Core Recipe" tag to the Preparations field. The feature only appends this tag, if the tag is not previously there.

## Test Plan
For the manual test plan, we want to use the feature to create a fresh new set of base recipe items that with the "Core Recipe" Preparations tag
1. Sign in to Galley [here](https://staging-app.galleysolutions.com/login)
2. Go to [Preparation: Core Recipe](https://staging-app.galleysolutions.com/preparations/cHJlcGFyYXRpb246MzEzNjk=) of the Galley staging-app to view all recipes that contain a "Core Recipe" preparation

### Testing a single recipe id:
3. Open terminal and run `python` from the `galley-sdk` root folder:
```python3
>>> from galley.queries import *; from galley.enums import *; from galley.mutations import *; from galley.management import *;

>>> get_untagged_core_recipe_item_ids(["cmVjaXBlOjIwMjI5NA=="])    # recipe id for Harissa Roasted Carrot Salad BASE
```
- [x]  Verify result: `['cmVjaXBlSXRlbToxMjkwMTgx', 'cmVjaXBlSXRlbToxMjkwMjAw']`

    - Note: These two recipe item ids correspond to the [Harissa Roasted Carrot Salad BASE / "cmVjaXBlOjIwMjI5NA=="](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjIwMjI5NA==) as a unique component of its two parent recipes (scroll to bottom of the page under the "USAGES" section):
        - (1) Harissa Roasted Carrot Salad with Paprika Chicken
        - (2) Harissa Roasted Carrot Salad with Paprika Kidney Beans

5. Run the batch job for adding Core Recipe preparation to all the recipe items that require the **Harissa Roasted Carrot Salad BASE** recipe above:  
```python3
>>> update_recipe_item_data({"ids": ["cmVjaXBlOjIwMjI5NA=="], "attrs": {"preparationIds": ["cHJlcGFyYXRpb246MzEzNjk="]}})
```
- Note: `update_recipe_item_data()` automatically runs `get_untagged_core_recipe_item_ids()` when called.

6.  Refresh page from Step 2 and verify that these recipes are shown to now contain a "Core Recipe" preparation:
    - [x] (1) Harissa Roasted Carrot Salad with Paprika Chicken
    - [ ] (2) Harissa Roasted Carrot Salad with Paprika Kidney Beans

### Testing multiple recipe ids:
7. Run method with three recipe ids passed in:   
```python3
>>> get_untagged_core_recipe_item_ids(["cmVjaXBlOjIwMjI5NA==", "cmVjaXBlOjE3NjQxNA==", "cmVjaXBlOjE4MDAwMw=="])    
# recipe ids for Harissa Roasted Carrot Salad BASE, Daikon Sunflower Salad BASE, & Roasted Apple & Squash Chipotle Chili BASE
```
- [x] Verify result: `['cmVjaXBlSXRlbToxMTUwNDY5', 'cmVjaXBlSXRlbToxMTUwNDE3', 'cmVjaXBlSXRlbToxMTY0Njg5', 'cmVjaXBlSXRlbToxMTY0NTgy', 'cmVjaXBlSXRlbToxMjkyNjU3', 'cmVjaXBlSXRlbToxMjkyNjYy']`
    
    - Note: The ids from Step 3 should not be included in this return list since they've already had the "Core Recipe" preparation applied in Step 5

9. Run the batch job for adding "Core Recipe" preparation to all recipe items that require the Harissa Roasted Carrot Salad BASE,  Daikon Sunflower Salad BASE, & Roasted Apple & Squash Chipotle Chili BASE. 
```python3
update_recipe_item_data({"ids": ["cmVjaXBlOjIwMjI5NA==", "cmVjaXBlOjE3NjQxNA==", "cmVjaXBlOjE4MDAwMw=="], "attrs": {"preparationIds": ["cHJlcGFyYXRpb246MzEzNjk="]}})
```
10. Refresh page from Step 2 and verify that these recipes are shown to now contain a "Core Recipe" preparation:
    - 2 usage items for [Harissa Roasted Carrot Salad BASE / cmVjaXBlOjIwMjI5NA==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjIwMjI5NA==)
        - [x] (1) Harissa Roasted Carrot Salad with Paprika Chicken
        - [x] (2) Harissa Roasted Carrot Salad with Paprika Kidney Beans
    - 2 usage items for [Daikon Sunflower Salad BASE / cmVjaXBlOjE3NjQxNA==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE3NjQxNA==)
        - [x] (1) Daikon Sunflower Salad with Ginger Maple Dressing and Garlic Chicken
        - [x] (2) Daikon Sunflower Salad with Ginger Maple Dressing and Garlic Mung Beans
    - 4 usage items for [Roasted Apple & Squash Chipotle Chili BASE / cmVjaXBlOjE4MDAwMw==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE4MDAwMw==)
        - [x] (1) Roasted Apple & Squash Chipotle Chili with Herb Pinto Beans
        - [x] (2) Roasted Apple & Squash Chipotle Chili with Paprika Chicken
        - [x] (3) Roasted Apple & Squash Chipotle Chili with Savory Ground Turkey
        - [x] (4) Roasted Apple & Squash Chipotle Chili with Savory Pinto Beans
11. Notice that we've now ran the command `update_recipe_item_data()` twice on [Harissa Roasted Carrot Salad BASE / cmVjaXBlOjIwMjI5NA==](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjIwMjI5NA==). Go to parent recipe pages and verify that **ONLY ONE** "Core Recipe" preparation chip is present for the Harissa Roasted Carrot Salad BASE recipe under the "COMPONENTS" section:
    - [x] (1) [Harissa Roasted Carrot Salad with Paprika Chicken](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjIwMjMwNQ==)
    - [x] (2) [Harissa Roasted Carrot Salad with Paprika Kidney Beans](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjIwMjMwMA==)

### Automated Test
```python -m unittest tests/test_*```

### DIFF
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -1,13 +1,24 @@
 from sgqlc.operation import Operation
 from sgqlc.types import Field, Type
-
+from typing import Dict, List, Optional, Union, Any
 from galley.common import make_request_to_galley, validate_response_data
-from galley.types import (BulkMenusInput, MenuInput, MenuItemInput, MenuPayload, UnitInput)
+from galley.queries import get_recipe_item_ids
+from galley.types import (
+    BulkMenusInput,
+    MenuInput,
+    MenuItemInput,
+    MenuPayload,
+    UnitInput,
+    BulkUpdateRecipeItemsPayload,
+    BulkUpdateRecipeItemsInput,
+    RecipeItemInput
+)
 
 
 # This is graphql root for mutating data according to sgqlc lib. So this class name has to be Mutation.
 class Mutation(Type):
     bulkUpsertMenus = Field(MenuPayload, args={'input': BulkMenusInput})
+    bulkUpdateRecipeItems = Field(BulkUpdateRecipeItemsPayload, args={'input': BulkUpdateRecipeItemsInput})
 
 
 # MENU MUTATIONS
@@ -20,8 +31,8 @@ def build_unit_input(item):
         name = item['unit_name']
     else:
         name = 'each'
-
     return UnitInput(name=name)
+
 
 def build_menu_item_inputs(items):
     menu_item_inputs = []
@@ -35,6 +46,7 @@ def build_menu_item_inputs(items):
         menu_item_inputs.append(menuItemInput)
     return menu_item_inputs
 
+
 def build_menu_inputs(menus):
     menu_inputs = []
     for menu in menus:
@@ -44,8 +56,8 @@ def build_menu_inputs(menus):
         menu_input = MenuInput(id=menu['menu_id'])
         menu_input.menuItems = build_menu_item_inputs(menu['menu_items'])
         menu_inputs.append(menu_input)
-
     return menu_inputs
+
 
 def build_upsert_mutation_query(args):
     mutation = Operation(Mutation)
@@ -53,6 +65,7 @@ def build_upsert_mutation_query(args):
     bulk_input.menus = build_menu_inputs(args)
     mutation.bulkUpsertMenus(input=bulk_input)
     return mutation
+
 
 # args = [
 # {
@@ -67,5 +80,38 @@ def build_upsert_mutation_query(args):
 # ]
 def upsert_menu_data(args):
     mutation = build_upsert_mutation_query(args)
+    response = make_request_to_galley(op=mutation)
+    return validate_response_data(response)
+
+
+def build_update_mutation_query(args):
+    ids = []
+    if args["ids"] is not None and len(args["ids"]) > 0:
+        ids = {idx for idx in args["ids"] if type(idx).__name__ == "str"}
+        if len(ids) <= 0:
+            raise ValueError("non valid ID provided, all IDs must be strings")
+
+    mutation = Operation(Mutation)
+    bulk_input = BulkUpdateRecipeItemsInput(
+        ids=ids,
+        attrs=RecipeItemInput(args["attrs"])
+    )
+    mutation.bulkUpdateRecipeItems(input=bulk_input)
+    return mutation
+
+
+def update_recipe_item_data(args):
+    if "attrs" not in args.keys() or args["attrs"] is None:
+        raise ValueError("attrs property not provided")
+
+    if "ids" not in args.keys() or args["ids"] is None or len(args["ids"]) <= 0:
+        raise ValueError("recipe item id list not provided")
+
+    recipe_item_ids = get_recipe_item_ids(args["ids"])
+    payload = {
+        "ids": recipe_item_ids,
+        "attrs": args["attrs"],
+    }
+    mutation = build_update_mutation_query(payload)
     response = make_request_to_galley(op=mutation)
     return validate_response_data(response)
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -110,10 +110,14 @@ def build_bulk_update_recipe_item_query(args):
 
 def bulk_update_recipe_item_data(args):
     if not args.get("attrs"):
-        raise ValueError("attrs property not provided")
+        msg = f"{GALLEY_ERROR_PREFIX} attrs property not provided"
+        logger.exception(msg)
+        raise ValueError(msg)
 
     if not args.get("ids"):
-        raise ValueError("recipe item id list not provided")
+        msg = f"{GALLEY_ERROR_PREFIX} Recipe item id list not provided"
+        logger.exception(msg)
+        raise ValueError(msg)
 
     # apply filter to ensure that only untagged recipe items are selected for tagging
     recipe_item_ids = get_untagged_core_recipe_item_ids(args["ids"])
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -13,6 +13,7 @@ from galley.types import (
     BulkUpdateRecipeItemsInput,
     RecipeItemInput
 )
+from galley.enums import PreparationEnum
 
 
 # This is graphql root for mutating data according to sgqlc lib. So this class name has to be Mutation.
@@ -107,7 +108,12 @@ def update_recipe_item_data(args):
     if "ids" not in args.keys() or args["ids"] is None or len(args["ids"]) <= 0:
         raise ValueError("recipe item id list not provided")
 
-    recipe_item_ids = get_recipe_item_ids(args["ids"])
+    filters = [{
+        "haystack_key": "id",
+        "needle": PreparationEnum.CORE_RECIPE.value,
+        "isFalse": True
+    }]
+    recipe_item_ids = get_recipe_item_ids(args["ids"], filters)
     payload = {
         "ids": recipe_item_ids,
         "attrs": args["attrs"],
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -123,9 +123,6 @@ def bulk_update_recipe_item_data(args):
     mutation = build_bulk_update_recipe_item_query(payload)
     response = make_request_to_galley(op=mutation)
 
-    from pprint import pprint
-    pprint(response)
-
     valid_response = validate_response_data(response)
 
     if valid_response is None:
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -34,7 +34,6 @@ def build_unit_input(item):
         name = 'each'
     return UnitInput(name=name)
 
-
 def build_menu_item_inputs(items):
     menu_item_inputs = []
     for item in items:
@@ -47,7 +46,6 @@ def build_menu_item_inputs(items):
         menu_item_inputs.append(menuItemInput)
     return menu_item_inputs
 
-
 def build_menu_inputs(menus):
     menu_inputs = []
     for menu in menus:
@@ -59,14 +57,12 @@ def build_menu_inputs(menus):
         menu_inputs.append(menu_input)
     return menu_inputs
 
-
 def build_upsert_mutation_query(args):
     mutation = Operation(Mutation)
     bulk_input = BulkMenusInput()
     bulk_input.menus = build_menu_inputs(args)
     mutation.bulkUpsertMenus(input=bulk_input)
     return mutation
-
 
 # args = [
 # {
@@ -84,13 +80,12 @@ def upsert_menu_data(args):
     response = make_request_to_galley(op=mutation)
     return validate_response_data(response)
 
-
-def build_update_mutation_query(args):
+def build_bulk_update_recipe_item_query(args):
     ids = []
     if args.get("ids"):
         ids = {id for id in args["ids"] if type(id) == str}
         if len(ids) <= 0:
-            raise ValueError("non valid ID provided, all IDs must be strings")
+            raise ValueError("No valid IDs provided. All IDs must be strings.")
 
     mutation = Operation(Mutation)
     bulk_input = BulkUpdateRecipeItemsInput(
@@ -99,7 +94,6 @@ def build_update_mutation_query(args):
     )
     mutation.bulkUpdateRecipeItems(input=bulk_input)
     return mutation
-
 
 def update_recipe_item_data(args):
     if not args.get("attrs"):
@@ -114,6 +108,6 @@ def update_recipe_item_data(args):
         "ids": recipe_item_ids,
         "attrs": args["attrs"],
     }
-    mutation = build_update_mutation_query(payload)
+    mutation = build_bulk_update_recipe_item_query(payload)
     response = make_request_to_galley(op=mutation)
     return validate_response_data(response)
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -87,9 +87,12 @@ def upsert_menu_data(args):
 
 def build_bulk_update_recipe_item_query(args):
     ids = []
+    invalid_ids = []
     if args.get("ids"):
-        ids = {id for id in args["ids"] if type(id) == str}
-        invalid_ids = [id for id in args["ids"] if id not in ids]
+        for id in args["ids"]:
+            ids.append(id)
+        else:
+            invalid_ids.append(id)
         for invalid_id in invalid_ids:
             logger.info(f"{GALLEY_ERROR_PREFIX} Key {invalid_id} \
                          is not a valid id and will not be included \
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -2,7 +2,7 @@ from sgqlc.operation import Operation
 from sgqlc.types import Field, Type
 from typing import Dict, List, Optional, Union, Any
 from galley.common import make_request_to_galley, validate_response_data
-from galley.queries import get_untagged_core_recipe_item_ids
+from galley.queries import get_untagged_core_recipe_item_ids_via_connection
 from galley.types import (
     BulkMenusInput,
     MenuInput,
@@ -115,7 +115,7 @@ def bulk_update_recipe_item_data(args):
         raise ValueError(msg)
 
     # apply filter to ensure that only untagged recipe items are selected for tagging
-    recipe_item_ids = get_untagged_core_recipe_item_ids(args["ids"])
+    recipe_item_ids = get_untagged_core_recipe_item_ids_via_connection(args["ids"])
     payload = {
         "ids": recipe_item_ids,
         "attrs": args["attrs"],
--- a/galley/mutations.py
+++ b/galley/mutations.py
@@ -87,16 +87,8 @@ def upsert_menu_data(args):
 
 def build_bulk_update_recipe_item_query(args):
     ids = []
-    invalid_ids = []
     if args.get("ids"):
-        for id in args["ids"]:
-            ids.append(id)
-        else:
-            invalid_ids.append(id)
-        for invalid_id in invalid_ids:
-            logger.info(f"{GALLEY_ERROR_PREFIX} Key {invalid_id} \
-                         is not a valid id and will not be included \
-                         in the query")
+        ids = [id for id in args["ids"] if type(id) == str]
         if len(ids) <= 0:
             msg = f"{GALLEY_ERROR_PREFIX} No valid IDs provided. \
                     All IDs must be strings."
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: galley/mutations.py [hunk 1]
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
     mutation = build_upsert_mutation_query(args)
     response = make_request_to_galley(op=mutation)
     return validate_response_data(response)
+
+
+def build_update_mutation_query(args):
+    ids = []
+    if args["ids"] is not None and len(args["ids"]) > 0:
```
thunguy, line 90:
```suggestion
    if args.get("ids"): 
```

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+                variables={'id': recipe_ids}),
+            'recipes')
+    return validated_response_data
+
+
+def get_untagged_core_recipe_item_ids(ids):
+    recipe_item_ids = []
+    throw_error = True
+    if ids is not None and len(ids) > 0:
+        ids = [id for id in ids if type(id).__name__ == "str"]
```
thunguy, line 250:
```suggestion
        ids = [id for id in ids if type(id) == str]
```

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+                op=query.__to_graphql__(auto_select_depth=3),
+                variables={'id': recipe_ids}),
+            'recipes')
+    return validated_response_data
+
+
+def get_untagged_core_recipe_item_ids(ids):
+    recipe_item_ids = []
+    throw_error = True
+    if ids is not None and len(ids) > 0:
```
thunguy, line 249:
```suggestion
    if ids:
```
This simplified line will check for both cases.

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+        ids = [id for id in ids if type(id).__name__ == "str"]
+        throw_error = len(ids) <= 0
+        if throw_error is False:
```
thunguy, line 252:
```suggestion
        if not throw_error:
```

thunguy: More "pythonic" way to state this.

galley/mutations.py
thunguy, line 90:
```suggestion
    if args.get("ids"):
```
This will check for both cases. Also, if "ids" isn't actually in args, `args["ids"]` would still throw an error here so its best to use `.get()` for checking.

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+        ids = {idx for idx in args["ids"] if type(idx).__name__ == "str"}
```
thunguy, line 91:
```suggestion
        ids = {id for id in args["ids"] if type(id) == str}
```

thunguy: Non-blocking on the variable name but just want to make sure we're being clear about the type of item we're evaluating here--I normally associate 'idx' to mean 'index', but in this case each item being iterated over is an id.

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+    bulk_input = BulkUpdateRecipeItemsInput(
+        ids=ids,
+        attrs=RecipeItemInput(args["attrs"])
+    )
+    mutation.bulkUpdateRecipeItems(input=bulk_input)
+    return mutation
+
+
+def update_recipe_item_data(args):
+    if "attrs" not in args.keys() or args["attrs"] is None:
```
thunguy, line 105:
```suggestion
    if not args.get("attrs"):
```

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+        raise ValueError("attrs property not provided")
+
+    if "ids" not in args.keys() or args["ids"] is None or len(args["ids"]) <= 0:
```
thunguy, line 108:
```suggestion
    if not args.get("ids"):
```
This will check for both cases.

galley/mutations.py
thunguy: This will check for both conditions.

galley/mutations.py
thunguy: This will check for all three conditions.

galley/mutations.py
thunguy, line 91:
```suggestion
    if args.get("ids"):
        ids = {id for id in args["ids"] if type(id) == str}
```

[90] This will check for both cases. Also, if "ids" isn't actually in args, args["ids"] would still throw a KeyError here, so its best to use .get() for performing this check.

[91] Non-blocking on the variable name but just want to make sure we're being clear about the type of item we're evaluating here--I normally associate 'idx' to mean 'index', but in this case each item being iterated over is an id. Also, just a shorter way to check for type str.

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+                for id, recipe in zip(sorted(ids), recipes):
+                    for parent in recipe['parentRecipeItems']:
+                        for recipe_item in parent['recipe']['recipeItems']:
+                            if recipe_item['subRecipe']:
+                                if recipe_item['subRecipe']['id'] == id \
+                                    and PreparationEnum.CORE_RECIPE.value \
+                                        not in {prep.get('id') for prep in recipe_item['preparations']}:
+                                    recipe_item_ids.append(recipe_item['id'])
+    if throw_error:
+        raise ValueError("no valid recipe ids provided, all ids must in of string")
```
thunguy, line 265:
```suggestion
    error = True
    if ids:
        ids = [id for id in ids if type(id) == str]
        error = len(ids) <= 0
        if not error:
            recipes = get_raw_recipe_items_data(ids)
            if recipes:
                recipes = sorted(recipes, key=lambda r: r['id'])
                for id, recipe in zip(sorted(ids), recipes):
                    for parent in recipe['parentRecipeItems']:
                        for recipe_item in parent['recipe']['recipeItems']:
                            if recipe_item['subRecipe']:
                                if recipe_item['subRecipe']['id'] == id \
                                    and PreparationEnum.CORE_RECIPE.value \
                                        not in {prep.get('id') for prep in recipe_item['preparations']}:
                                    recipe_item_ids.append(recipe_item['id'])
    if error:
        raise ValueError("No valid recipe ids provided. All ids must be a string.")
```

[249] This simplified line will check for both cases.
[250] A shorter way to implement a check for str type.
[252] More "pythonic" way to check for a False / false-y condition.
[265] Raise a clearer error message about ids needing to be passed in as strings.

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+                op=query.__to_graphql__(auto_select_depth=3),
+                variables={'id': recipe_ids}),
+            'recipes')
+    return validated_response_data
+
+
+def get_untagged_core_recipe_item_ids(ids):
+    recipe_item_ids = []
+    error = True
+    if ids:
```
patrick-thistle, line 249: It would be great for readability of this function to pull out the input validation/exception raising as a separate method.

Check the ids, raise the exception if needed then carry on with the logic.

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+        ids = [id for id in ids if type(id) == str]
+        error = len(ids) <= 0
+        if not error:
+            recipes = get_raw_recipe_items_data(ids)
+            if recipes:
+                recipes = sorted(recipes, key=lambda r: r['id'])
+                for id, recipe in zip(sorted(ids), recipes):
```
patrick-thistle, line 256: I'm not a huge fan of the heavy reliance on order within the recipes / ids here... I think it invites potential edge cases,  if for instance we pass 3 ids into this function and one of the ids (lets say #2) doesn't return any corresponding recipe.  We're left in a position where we have 3 ids and 2 recipes and zip could potentially link the incorrect id to a block of recipe data.

A key value pair here feels like an easier, less clever and clearer solution that removes any requirement for order at all.  It would also have the benefit of potentially removing a couple of sort loops from an already loopy method.

The { 'ID': [RECIPE_DATA], 'ID2': [] } shape would mean if an id has no recipe info it gets set to an empty list and we move on with our days.

patrick-thistle:
It came up in conversation with Ugo today that the desire for this is to have as a one off for ops to run to identify recipes that need updating, not for regular execution.

If this is the case it might be good to look at creating a separate place to keep logic such as this where it's clearly documented and understood that these methods were built for a very specific purpose.  Potentially something like a management.py file??

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
     mutation = build_upsert_mutation_query(args)
     response = make_request_to_galley(op=mutation)
     return validate_response_data(response)
+
+
+def build_update_mutation_query(args):
```
smunro-thistle, line 88: - [ ] I would name this something like build_bulk_update_recipe_item_query so it's clear it is specific to Recipe Items (and specifically a bulk update!)

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+    ids = []
+    if args.get("ids"):
+        ids = {id for id in args["ids"] if type(id) == str}
```
smunro-thistle, line 91: - [ ] For a one-off job like this, I think logging is very helpful. I'd log each id that is invalid (or store in a separate list and then log that list) so we know which ones we didn't update
- [ ] I agree with @patrick-thistle that since this mutation is meant to be run manually for one-off bulk updates, we should move it to another a separate file.

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+        if len(ids) <= 0:
+            raise ValueError("non valid ID provided, all IDs must be strings")
```
smunro-thistle, line 93: - [ ] I'd make this error something like "No valid IDs provided. All IDs must be strings." The small change in wording means this error would make sense both if NO ids were provided, and if no VALID ids were provided.

galley/mutations.py
```diff
@@ -69,3 +83,37 @@ def upsert_menu_data(args):
...
+    mutation = Operation(Mutation)
+    bulk_input = BulkUpdateRecipeItemsInput(
+        ids=ids,
+        attrs=RecipeItemInput(args["attrs"])
+    )
+    mutation.bulkUpdateRecipeItems(input=bulk_input)
+    return mutation
+
+
+def update_recipe_item_data(args):
```
smunro-thistle, line 104: - [ ] I'd name this something like bulk_update_recipe_item_data since this is specifically for a bulk attribute update.
- [ ] Also move this to a different file with the method above

galley/queries.py
smunro-thistle: ^ Agree with this

galley/queries.py
```diff
@@ -212,3 +212,55 @@ def get_raw_menu_data(dates: List[str],
...
+                    for parent in recipe['parentRecipeItems']:
```
smunro-thistle, line 257: These loops are safer if you structure them like this (so you don't error out if one element doesn't contain the child element):

`for parent in recipe.get('parentRecipeItems', [])`

smunro-thistle (commented):
I agree with Patrick M's suggestion of pulling this out into a separate file (for example, management.py or scripts or something similar). Also some additional suggestions/changes. I haven't run through the test plan yet. I'll wait to do that until the code has been pulled out into another file.

galley/mutations.py
ugonwigwethistle: * The second approach for logging was employed, made a list of all invalid IDs and logged them.
* Management  / one-time scripts to be moved in a future commit

galley/queries.py
ugonwigwethistle: The convoluted loop-in-loop structure has been removed and replaced with a couple of double-loops, one for pre-preparing the recipeItem data and the second for searching the prepared recipeItem data.

Also this removes the zip() and sorted() functions as well, the pre-prep loop ensures that all entered IDs have either a valid recipe list (if available) or an empty list (if not)

galley/queries.py
```diff
@@ -212,3 +219,107 @@ def get_raw_menu_data(dates: List[str],
...
+def get_raw_recipe_items_data_via_connection(sub_recipe_ids: List) -> Iterable[List[Dict]]:
+    query = get_ops_recipe_item_connection_query(sub_recipe_ids=sub_recipe_ids)
+    validated_response_data = validate_response_data(
+            make_request_to_galley(
+                op=query.__to_graphql__(auto_select_depth=3),
+                variables={'subRecipeIds': sub_recipe_ids}),
+            'recipeItemConnection')
+    return validated_response_data
+
+def get_untagged_core_recipe_item_ids(ids):
```
smunro-thistle, line 267: Old to be deleted

galley/queries.py
```diff
@@ -212,3 +219,107 @@ def get_raw_menu_data(dates: List[str],
...
+
+def get_ops_recipe_item_connection_query(sub_recipe_ids: List[str]) -> Operation:
+    query = Operation(Query)
+    query.viewer.recipeItemConnection(filters=RecipeItemConnectionFilter(subRecipeIds=sub_recipe_ids)).__fields__('edges')
+    query.viewer.recipeItemConnection.edges.__fields__('node')
+    query.viewer.recipeItemConnection.edges.node.__fields__('id', 'recipeId', 'preparations')
+    query.viewer.recipeItemConnection.edges.node.preparations.__fields__('id', 'name')
+    return query
+
+def get_raw_recipe_items_data(recipe_ids: List) -> Iterable[List[Dict]]:
```
smunro-thistle, line 241: Old one to be removed?

galley/queries.py
```diff
@@ -212,3 +219,107 @@ def get_raw_menu_data(dates: List[str],
                     else:
                         continue
     return response
+
+def get_ops_recipe_items_query(recipe_ids: List[str]) -> Operation:
```
smunro-thistle, line 223: Are you going to remove this?

galley/queries.py
```diff
@@ -212,3 +219,107 @@ def get_raw_menu_data(dates: List[str],
...
+def get_untagged_core_recipe_item_ids_via_connection(ids):
+    recipe_item_ids = []
+    ids = [id for id in ids if type(id) == str]
+    if len(ids) <= 0:
+        error = "No valid recipe ids provided. All ids must be a string."
+        logger.exception(error)
+        raise ValueError(error)
+    recipe_item_connection = get_raw_recipe_items_data_via_connection(ids) or {}
+    for recipe_item in recipe_item_connection.get("edges", []):
+        recipe_item_ids.append(recipe_item["node"]["id"])
```
smunro-thistle, line 324: - [ ] Do you need to filter out the ones that are already tagged?

setup.py
```diff
@@ -4,7 +4,7 @@
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.35.2',
+    version='0.36.3',
```
smunro-thistle, line 7: I think the current version is now 36.0, so this now needs to be bumped to 37.0

tests/test_mutations.py
```diff
@@ -210,3 +213,181 @@ def test_exception_raised_if_no_menu_id_present(self):
...
+            "attrs": {
+                "preparationIds": ["cHJlcGFyYXRpb246MzEzNjk="]
+            }
+        }
+        ret = build_bulk_update_recipe_item_query(payload)
+        mutation_str = bytes(ret).decode("utf-8")
+        self.assertEqual(mutation_str, expected_str)
+
+    @mock.patch('galley.mutations.make_request_to_galley')
+    def test_bulk_update_recipe_item_data_raises_exception_if_thrown_from_request(self, mock_mr):
```
smunro-thistle, line 367: This test doesn't look like it actually tests that an exception is thrown?

tests/test_ops_queries.py
```diff
@@ -220,3 +222,85 @@ def test_get_ops_menu_query(self):
...
+            name
+            }
+            }
+            }
+            }
+            }
+            }'''.replace(' '*12, '')
+
+    def response(self, *data):
+        '''return ({
```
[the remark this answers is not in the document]
ugonwigwethistle: Yes!

tests/test_mutations.py
```diff
@@ -210,3 +213,181 @@ def test_exception_raised_if_no_menu_id_present(self):
...
+
+    def test_exception_raised_if_no_attrs_provided(self):
+        payload = {
+            "ids": ["cmVjaXBlOjIwMjI5NA==", None, 1],
+            "attrs": None
+        }
+        with self.assertRaises(ValueError):
+            bulk_update_recipe_item_data(payload)
+
+    def test_no_mutations_done_if_already_tagged_recipe_item_id_is_provided(self):
```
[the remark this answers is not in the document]
ugonwigwethistle: This test is to be removed as it performs a mutation (`build_bulk_update_recipe_item_query()`) even if the recipe items have been tagged. 

The actual parent mutation method `bulk_update_recipe_item_data()` first checks for duplications and filters out the tagged items before this tested method is called called. 

This test will be removed and another method to mock double calls with the parent method will be added.

tests/test_mutations.py
ugonwigwethistle: Check for exception updated in next commit.

smunro-thistle (approved):
There are few comments about deleting lines to take care of before merging!

PR #57, opened 2022-06-21
Replace ops query quantityUnitValue field with quantity & unit

thunguy:
## Description
This PR replaces the `quantityUnitValue` field from the ops menu query with `quantity` and `unit` to pull the quantities shown for recipe items in Galley. These are the quantity usages that our ops teams need, and knowing that, we are able to scope down the quantities needed to be returned by the query.

## Test Plan
## Versioning


ugonwigwethistle (approved):
Approved

PR #58, opened 2022-06-23
Add quantity unit values to ops menu query for component target amount

thunguy:
## Description
This PR reinstates the `quantityUnitValue` field at the first and second levels of `recipeTreeComponents`. By ridding this data, we lost the ability to calculate `component.target_amount`. Although ounces is the primary quantity usage in relation to plating meals,  staging needs these values in pounds. 

### Run Manual Tests
```python3
python

>>> from galley.queries import *; from galley.formatted_ops_queries import *; import json;
>>> print(json.dumps(get_formatted_ops_menu_data(["2022-06-20"]), indent=4))
```

1. Check that the formatted data returned by the query includes a `usage` and `quantity` object. Usage units will vary (mainly ounces), but all quantity objects should have unit equal to pounds.

### Run Automated Tests
```python3
python -m unittest tests/test_*
```

## Versioning
- [x] Please update the project version in setup.py (v0.35.1 -> v0.35.2)


ugonwigwethistle (approved):
Approved

PR #59, opened 2022-07-06
(VEG-88) adding version to recipe

talkallday:
## Description
adding `version` to response for `get_formatted_recipes_data`
## Test Plan
automated tests and to test manually, since version is not available in staging, 
1. log in to galley staging dashboard
2. select a recipe
3. edit something (easy is the name) to generate a version. (feel free to edit several times if you want to test pagination in query when there are many versions)
4. open shell
5. `>>> from galley.formatted_queries import get_formatted_recipes_data`
6.  confirm `version` is in the output of `>>> get_formatted_recipes_data(['<recipeID_for_recipe_with_versions_you_created'])`

## Versioning


ErikBrown (approved):
Tested locally, am able to get version numbers for edited staging recipes, and `None` for unedited recipes.

PR #62, opened 2022-09-14
Add ounce conversion to formatted menu ops

thunguy:
## Description
- [x]  Expose the oz quantity unit value to send with `get_formatted_ops_menu_data` to predominately get the oz conversion equivalent for any "each" quantity value usage.

## Test Plan
```python
from galley.formatted_ops_queries import *; import json
print(json.dumps(get_formatted_ops_menu_data(["2022-09-12"]), indent=4))

# check that all primary and secondary components contain a quantityValues list consisting of lb. and oz. values (if exists); tertiary components will only have a usage object, but not quantityValues
```

## Versioning

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.38.0',
+    version='0.38.1',
     packages=['galley'],
     install_requires=['sgqlc==14.0', 'backoff==1.11.1']
 )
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -4,7 +4,7 @@
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.38.0',
+    version='0.38.1',
```
smunro-thistle: I might make this 0.39, just because it's a breaking change for Okra. Thoughts?

smunro-thistle (approved):
Looks good!

PR #65, opened 2022-10-05
(FRU-298) Formatted ops menu returns cupping container data

thunguy:
## Description
get_formatted_ops_menu_data() returns 

## Test Plan

## Versioning


galley/formatted_ops_queries.py
```diff
@@ -166,33 +170,53 @@ def format_quantity_value(quantity_values: List) -> List[Dict]:
...
 
 
-def is_packaging(component: Dict) -> bool:
+def is_packaging(ingredient: Dict) -> bool:
     """
     Checks and returns True if an ingredient-typed component is a food
     package item.
     """
-    cvs = component.get('ingredient', {}).get('categoryValues') or []
+    cvs = ingredient.get('categoryValues') or []
```
smunro-thistle, line 187: NB: .get lets you default like this: `cvs = ingredient.get('categoryValues', [])`

galley/formatted_ops_queries.py
```diff
@@ -166,33 +170,53 @@ def format_quantity_value(quantity_values: List) -> List[Dict]:
...
     return any(cv.get('id') == IngredientCValEnum.FOOD_PACKAGE.value for cv in cvs)
 
 
+def get_cupping_container(preparations: List) -> Optional[bool]:
+    """
+    Returns True if a recipe component contains a cupping container
+    preparation.
+    """
+    containers = {
```
smunro-thistle, line 196: NB: Sorry if this is a duplicate comment, github isn't showing me the comment I added here! I'd be tempted to make this a constant of some kind, but we can always pull it out as a constant if we have other uses for it later

galley/formatted_ops_queries.py
```diff
@@ -166,33 +170,53 @@ def format_quantity_value(quantity_values: List) -> List[Dict]:
...
+                 for prep in preparations
+                 if prep.get('id') in containers), None)
+
+
 def format_ops_menu_rtc_data(recipe_tree_components: List) -> List[Optional[Dict]]:
     components: List = []
     for rtc in recipe_tree_components:
-        ingredient = rtc.get('ingredient') or {}
-        subrecipe = rtc.get('recipeItem', {}).get('subRecipe') or {}
+        recipeitem = rtc.get('recipeItem') or {}
```
smunro-thistle, line 213: NB: Same .get comment as before (and same below). Can simplify to `recipeitem = rtc.get('recipeItem', {})`

tests/test_ops_formatted_queries.py
```diff
@@ -131,20 +133,40 @@ def test_format_instructions_None(self):
         self.assertEqual(result, [])
 
 
+class TestFormattedCuppingContainerData(TestCase):
```
smunro-thistle, line 136: Do you want to add an additional test for a container nested in a Core recipe?

smunro-thistle (commented):
Just a few comments. Also I wanted to better understand the switch from .ingredient to recipeItem.ingredient in the query

tests/test_ops_formatted_queries.py
thunguy: I have this scenario mocked out in `mock_ops_menu.mock_recipeTreeComponents`, and the tests from `TestGetFormattedOpsMenuData` is capturing the data and the expected results correctly. Rather than making another mock call to the api, I added a simpler test to test the formatted primary recipe component result.

galley/formatted_ops_queries.py
thunguy: I actually do this on purpose to handle for both cases where the key does exists with a None value, and if the key doesn't exist.

galley/formatted_ops_queries.py
thunguy: I actually do this on purpose to handle for both cases where the key does exists with a None value, and if the key doesn't exist.

galley/formatted_ops_queries.py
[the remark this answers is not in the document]
thunguy: Created a new ContainerEnum class to separate out cupping containers. Curious what your thoughts are!

thunguy:
@smunro-thistle  Going from ingredient to recipeItem.ingredient allows one less field to be brought in under recipeTreeComponents which I find it easier to parse and makes more sense given  how recipeItem is set up in Galley. Since recipeItem can only be either a subrecipe or an ingredient, never both, it makes it more convenient and takes less code to null check for either ingredient or subrecipe, rather than needing to grab the ingredient item and if not ingredient, then go into recipeItem and grab the subrecipe, when we could just do all that from the recipeItem level to begin with. It also makes more sense in regards to how preparations set up. Preparations live at the recipeItem level, so if a top level ingredient of a specific recipe has a preparation we care about, we can’t capture that from the ingredient object directly. So if we’re just looking at recipeItem which has preparations, usage amount, and quantity unit values data associated to either the ingredient or subrecipe, I think it makes more sense to bring in the context of the ingredient so that in the case the subrecipe is null, its clear as to what the actual recipeItem is.

PR #66, opened 2022-10-05
Added location to menu query

ErikBrown:
## Description
Adds location_id to the `get_menu_query` and `get_ops_menu_query` functions. There is a limit of 25 menus returned, and had been fetching ALL menus then filtering them down. This will cause issues once we activate the Burlington menus. This PR adds a filter to the GraphQL query itself so we don't need to return as much data.

## Test Plan
Test that `get_raw_menu_data` returns only the menus for the location provided to it. Here's some code to help with that testing:
### Thistle-web menu testing:
```
from galley.queries import get_raw_menu_data
expected_dates = ['2022-10-10', '2022-10-13', '2022-10-17', '2022-10-20', '2022-10-24', '2022-10-27', '2022-10-31', '2022-11-03', '2022-11-07', '2022-11-10']
raw_data_burlington = get_raw_menu_data(expected_dates, 'Burlington', 'production')

len(raw_data_burlington) # Should be 10
[[m['date'], m['location']['name']] for m in raw_data_burlington] # Can verify that it's only Burlington menus

raw_data_vacaville = get_raw_menu_data(expected_dates, 'Vacaville', 'production')

len(raw_data_vacaville) # Should be 10
[[m['date'], m['location']['name']] for m in raw_data_vacaville] # Can verify that it's only Vacaville menus
```
### Okra menu testing (takes a couple minutes to fetch):
```
raw_ops_data = get_raw_menu_data(expected_dates, 'Burlington', 'production', is_ops=True)

len(raw_ops_data) # Should be 10
[[m['date'], m['location']['name']] for m in raw_ops_data] # Can verify that it's only Burlington menus
```

**Test on BOTH staging and production Galley**

## Versioning

### DIFF
--- a/galley/enums.py
+++ b/galley/enums.py
@@ -1,6 +1,14 @@
 from enum import Enum
 
 
+class LocationEnum(Enum):
+    """
+    Enum for locations. <location name>: <location id>
+    """
+    VACAVILLE = 'bG9jYXRpb246MTkyOA=='
+    BURLINGTON = 'bG9jYXRpb246Mzg1MQ=='
+
+
 class MenuCategoryEnum(Enum):
     """
     Enum for categories, for item type menu <category name>: <category id>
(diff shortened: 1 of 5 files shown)


patrick-thistle (approved):
Everything tests out as expected.

Very excited to get this going in thistle-web!

PR #70, opened 2022-10-13
(FRU-278) Use ingredient name instead of externalName

smunro-thistle:
## Description
This updates ops queries to use ingredient names instead of externalNames in primary components.

## Test Plan
in a python console:
```
from galley.formatted_ops_queries import *
get_formatted_ops_menu_data(dates=['2022-10-31'])
```
Check that you see `'spring mix & arugula PF, SEND TO PLATE'` instead of `'Spring Mix Lettuce*, Baby Arugula*'` in the output

### DIFF
--- a/galley/formatted_ops_queries.py
+++ b/galley/formatted_ops_queries.py
@@ -57,7 +57,7 @@ class FormattedRecipeComponent:
         component = dict(
             type=self.type,
             id=self.data.get('id'),
-            name=get_external_name(self.data),
+            name=self.data.get('name'),
             usage=dict(value=self.quantity, unit=self.unit),
             quantityValues=format_quantity_value(self.quantity_values),
             binWeight=format_bin_weight(self.data.get('categoryValues')),
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 # are dependencies that are required on remote installs, not local development.
 setup(
     name='galley_sdk',
-    version='0.42.1',
+    version='0.42.2',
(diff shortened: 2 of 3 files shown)


galley/formatted_ops_queries.py
```diff
@@ -57,7 +57,7 @@ def to_primary_component_dict(self):
         component = dict(
             type=self.type,
             id=self.data.get('id'),
-            name=get_external_name(self.data),
+            name=self.data.get('name'),
```
jamie-smyth-at-thistle, line 60: Would there be any advantage to having both, incase it turns out we do need the external name at some point (I could see there being value in having both available)?

smunro-thistle: Hmm, yeah, that makes sense

smunro-thistle: Though we don't currently do that for subcomponents, so it would be a bigger lift to make them consistent (since I have to edit our mock data)

thunguy: I agree with both points, but because we've been informed that the `externalName` isn't necessary (see slack thread [here](https://thistle.slack.com/archives/C02CJA3QZ5H/p1663697241714859?thread_ts=1663695872.381479&cid=C02CJA3QZ5H)) for teams working in Okra, I think bringing in just the `name` field is reasonable. If ever we need to pull in `externalName`, the amount of work to do so would be a small lift.

thunguy (approved):
Data returns as expected. Good to go.

PR #71, opened 2022-10-26
VEG-261: make recipe details retrieval location aware

kate-skorija:
## Description
## Test Plan
1. In Galley staging, pull up a recipe like [Sesame Edamame Dip](https://staging-app.galleysolutions.com/recipes/cmVjaXBlOjE4MTY3Mg==?userLocationId=bG9jYXRpb246MTkyOA%3D%3D) that has base recipe, and a component with vendor items from multiple locations like [frozen edamame](https://staging-app.galleysolutions.com/ingredients/aW5ncmVkaWVudDoyNDQ0NDM=?userLocationId=bG9jYXRpb246MTkyOA%3D%3D)
2. Navigate to a location specific vendor item (West or East Coast), and under the nutrition tab, add unique dietary flags and nutrition data to the vendor item. For nutrition data it is helpful to pick a field that would be 0 otherwise, such as added sugar.
3. In the galley-sdk virtual environment, call `get_formatted_recipes_data` and `get_raw_recipes_data` with the recipe id and location name of the vendor item you modified. Example queries for Sesame Edamame Dip, Burlington:
```
import galley
from galley.queries import *

get_formatted_recipes_data(recipe_ids=['cmVjaXBlOjE4MTY3Mg=='], location_name='Burlington')
get_raw_recipes_data(recipe_ids=['cmVjaXBlOjE4MTY3Mg=='], location_name='Burlington')
```
4. Verify that your modifications from Galley are present in the results. 
 - For the `get_formatted_recipes_data` result, the dietary flags you added should be listed as values for `allergens`, and the nutritional info in `nutrition`.  
 - For `get_raw_recipes_data`, the nutrition information you modified should be present in the `reconciledNutritionals` at the top most level of the data, as well as in the `reconciledNutritionals` info for the base recipe within the recipe. `dietaryFlagsWithUsages` should also be present at the top most level with your additional flag ids. 
5. Repeat these calls with the opposite location name, and verify that the return data matches what is in Galley for that location. 
## Versioning
0.43.0


galley/queries.py
```diff
@@ -98,29 +100,36 @@ def recipe_connection_query(
...
     query.viewer.recipeConnection.edges.node.recipeTreeComponents.recipeItem.subRecipe.allIngredientsWithUsages.ingredient.\
         __fields__('id', 'externalName', 'name')
-    query.viewer.recipeConnection.edges.node.dietaryFlagsWithUsages.dietaryFlag.\
+    query.viewer.recipeConnection.edges.node.reconciledNutritionals(location_id=location_id)
+    query.viewer.recipeConnection.edges.node.dietaryFlagsWithUsages(location_id=location_id).dietaryFlag.\
         __fields__('id')
     return query
 
-def get_raw_recipes_data(recipe_ids: List[str]) -> Optional[List[Dict]]:
+def get_raw_recipes_data(recipe_ids: List[str], location_name: str='Vacaville') -> Optional[List[Dict]]:
```
thunguy, line 116: Would it make sense to pass in a list of locations instead? That way we make a single request for multiple locations rather than sending n requests for n locations we want to query for.

tests/test_queries.py
```diff
@@ -213,372 +213,21 @@ def test_get_raw_menu_data_filter_by_menu_type(self, mock_retrieval_method):
...
-            pageInfo {
-            endIndex
-            hasNextPage
-            hasPreviousPage
-            startIndex
-            }
-            }
-            }
-            }'''.replace(' '*12, '')
+    def test_recipe_connection_query_includes_reconciledNutritionals_with_locationId(self):
```
jamie-smyth-at-thistle, line 216: Everything in these changes to the test here is an improvement. Only testing one thing at a time.  Removing a massive setup block. Just lots of wonderful stuff here. ❤️

galley/queries.py
thunguy: Nevermind my above comment, I was thinking from the perspective of fetching menu data from multiple locations, which makes more sense then to pass in a list of locations. However, for specific recipes connected to one location, passing in a single argument makes sense.

patrick-thistle (changes requested):
Testing the location configuration with allergens and nutritional information worked great!

There was a last minute update to the AC around raising an exception vs. providing a default. 

Once branch is updated I will re-check and approve!

galley/formatted_queries.py
```diff
@@ -487,7 +488,7 @@ def get_category_menu_type(menu_category_values) -> str:
 
 def get_formatted_recipes_data(
     recipe_ids: List[str],
-    location_name: str='Vacaville',
+    location_name: str,
     **format_options: Dict
 ) -> Optional[List[Dict]]:
     recipes_data = get_raw_recipes_data(recipe_ids=recipe_ids, location_name=location_name) or []
```
patrick-thistle, line 494: can we add an explicit check / raise ArgumentError in location_name is either blank or None?

An exception will get raised currently if either is passed, but it gets thrown by the attempt to call .upper() in get_raw_recipes_data and it would be nice to have an explicit check here.

patrick-thistle (approved):
LGTM!  Nice work

PR #75, opened 2023-03-16
(ENG-1674) Refactor query for recipe_connection_query 

thunguy:
## Description
- Refactors `recipe_connection_query` and `get_formatted_recipes_data` that powers all recipe data thistle-web receives to reduce inefficiencies caused by querying nested `recipeTreeComponents`, and `quantityUnitValues`
- Query pulls in new fields `ingredientsWithUsages`, `usages.ancestorRecipes`, and `unit.UnitValues`

**Note:** Mock tests/data related to `recipeTreeComponents` is still currently being used by the ops menu data query, which will be refactored in a subsequent PR. For the changes made in this PR, any real/mock data related to `recipeTreeComponents` should be expunged. 

## Test Plan
1. Checkout main
2. Open python shell and run:
```python
from galley.formatted_queries import *; from pprint import pprint

formatted_menus = get_formatted_menu_data(['2023-04-10', '2023-04-13', '2023-04-17', '2023-04-20', '2023-04-24', '2023-04-27'], 'Vacaville')
formatted_recipes = get_formatted_recipes_data([mi['recipeId'] for m in formatted_menus for mi in m['menuItems']], 'Vacaville')

pprint(formatted_menus)
pprint(formatted_recipes)
```
3. Copy and paste output in a [diff checker](https://www.diffchecker.com/)
4. Checkout this branch
5. Repeat Step 2 and Step 3 to compare the diff

**Note:** If you're seeing slight mismatches between a few ingredient list diffs, the issue is most likely due to slight float precision differences caused from using `quantityUnitValues` (previously) and `unitValues` (currently) for ingredient calculations. You can verify the ingredient list for a recipe in Galley with their sort by usages feature, but keep in mind that Galley often rounds their ingredient usage values shown in the UI, and the difference should be a negligible amount. For context, our ingredient lists are primary sorted by usage in descending order, and then secondary sorted by external (common) name in ascending order, as required by the FDA. 

**Note:** Our "Ingredients" object returned by `get_formatted_recipes_data` does _not_ include standalone ingredients, so if using the Galley UI to compare "Ingredients" values at the _sellable_ recipe level, remember to factor out standalone usages.

- [ ] Verify `get_formatted_menu_data` in v0.45.0 returns the same shape and data values as it does in v0.44.0
- [ ] Verify `get_formatted_recipes_data` in v0.45.0 returns the same shape and data values as it does in v0.44.0

## Versioning
0.44.0 → 0.45.0


galley/formatted_queries.py
```diff
@@ -75,203 +83,132 @@ def calculate_serving_size_weight(
...
-    ):
-        self.subrecipe = subrecipe
-        self.nutrition = nutrition
-        self.ingredient = ingredient
-        self.unit_values = unit_values
-        self.preparations = preparations
-        self.category_values = ingredient.get('categoryValues', []) if ingredient else []
-        self.usages = subrecipe.get('allIngredientsWithUsages', []) if subrecipe else []
+    def __init__(self, recipe_item: Dict) -> None:
+        recipe_item = deepcopy(recipe_item)
```
ErikBrown, line 101: Why do we need to make a deepcopy here? Is it just for the mutating `pop`s on line 110 and 111?

thunguy: @ErikBrown That's correct.

galley/formatted_queries.py
```diff
@@ -283,25 +220,43 @@ def get_recipe_allergens(recipe_dietry_flags: List[Dict]) -> Dict:
         DietaryFlagEnum.SMOKED_MEATS.value: 'smoked meats',
         DietaryFlagEnum.BEEF.value: 'beef'
     }
-    allergens = []
+    allergens = [
+        allergen for df in dietry_flags
+        if (allergen:= allergen_labels.get(df.get('dietaryFlag', {}).get('id')))
```
ErikBrown, line 225: Adding a space around the assignment here
```suggestion
        if (allergen := allergen_labels.get(df.get('dietaryFlag', {}).get('id')))
```

ErikBrown, line 225: Is this meant to be an expression assignment here, or should this be `==`?

galley/formatted_queries.py
```diff
@@ -283,25 +220,43 @@ def get_recipe_allergens(recipe_dietry_flags: List[Dict]) -> Dict:
...
+def get_recipe_weights(recipe_items: List[Dict]) -> Dict:
+    """
+    Returns a dictionary containing total weight of recipe and
+    subrecipe details.
+    """
+    net_weight, gross_weight, label = 0, 0, None
+
+    for ri in recipe_items:
+        recipe_item = RecipeItem(ri)
+        label = l if (l:= recipe_item.get_label_name()) else label
```
ErikBrown, line 242:
```suggestion
        label = recipe_item.get_label_name() or label
```

galley/formatted_queries.py
```diff
@@ -312,29 +267,22 @@ def get_recipe_category_tags(
...
-        tag_id = recipe_category_value.get('category', {}).get('id')
-        label = recipe_tag_labels.get(tag_id)
-        recipe_category_value_name = recipe_category_value.get('name')
-
-        if label and recipe_category_value_name:
-            recipe_tags.setdefault(label, recipe_category_value_name)
-
+    recipe_tags = {
+        label: name for cv in category_values
+        if (label:= tag_labels.get(cv.get('category', {}).get('id'))) and (name:= cv.get('name'))
```
ErikBrown, line 272: NB: In general, it makes it easier to parse for me when the in-expression assignment is surrounded by spaces, since at first it looks an out of place malformed dictionary entry like this

galley/formatted_queries.py
```diff
@@ -363,99 +309,98 @@ def format_highlight_tags(
...
+                    f"standalone for recipe: {recipe_item.root}."
+                )
+
+    for iu in (recipe.get('ingredientsWithUsages') or []):
+        ingredient = RecipeItem(iu)
+
+        if not ingredient.is_packaging():
+            for usage in ingredient.usages:
+                ingredient_usage = dict(
+                    ingredient=ingredient.ingredient,
```
ErikBrown, line 334: NB: Spaces around =

galley/formatted_queries.py
ErikBrown: No, just for the variable assignment, not for the key assignment.
```Python
# In dict
{
  key: value
}

# In expression
(foo := 'bar')
```

galley/formatted_queries.py
thunguy: Good catch!

galley/formatted_queries.py
thunguy: Fixed all instances.

galley/formatted_queries.py
thunguy: Agreed! Fixed all instances for the assignment expression.

galley/formatted_queries.py
thunguy: Initially I just wanted to stick to one pattern for consistency sake, but I've actually updated the code to replace `dict()` with `{}` since initializing with the constructor is actually slower.

PR #77, opened 2023-03-27
(ENG-1675) Refactor menu ops query

thunguy:
## Description
- Refactors `get_ops_menu_query` and `get_formatted_ops_menu_data` that powers all menu/recipe data Okra receives to reduce inefficiencies caused by querying nested `recipeTreeComponents` and `quantityUnitValues`
- Query pulls in the entire flattened list of `recipeTreeComponents` and uses `ancestorComponentIds` to filter and build a DAG recipe mapping for plating and staging data
- Query replaces `quantityUnitValues` with `unit.UnitValues` for unit conversions

## Test Plan
1. Run a previous version of galley-sdk **before** v0.45.0
2. Open python shell and run:
```python
from galley.formatted_ops_queries import *; from pprint import pprint

pprint(get_formatted_ops_menu_data(['2023-04-10', '2023-04-13', '2023-04-17', '2023-04-20', '2023-04-24', '2023-04-27'], 'Vacaville'))
```
3. Copy and paste output in a [diff checker](https://www.diffchecker.com/)
4. Checkout this branch
5. Repeat Step 2 and Step 3 to compare the diff

- [x] Verify `get_formatted_ops_menu_data` in v0.46.0 returns the same shape and data values as it did prior to v0.45.0

## Versioning
0.45.0 → 0.46.0

### DIFF
--- a/tests/test_ops_formatted_queries.py
+++ b/tests/test_ops_formatted_queries.py
@@ -23,7 +23,7 @@ def formatted_ops_menu(date, location_name=DEFAULT_LOCATION, menu_type=DEFAULT_M
             'platePhotoUrl': 'https://cdn.filestackcontent.com/2X5ivrEYQvuEh30DyYot',
             'totalCount': 923,
             'totalCountUnit': 'each',
-            'primaryRecipeComponents': [],
+            'primaryRecipeComponents': MOCK_FORMATTED_PRIMARY_RECIPE_COMPONENTS,
 
         }, {
             'menuItemId': 'MENUITEM2DEF-OPS',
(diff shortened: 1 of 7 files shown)


tests/test_ops_formatted_queries.py
```diff
@@ -38,7 +23,7 @@ def formatted_ops_menu(date, location_name=DEFAULT_LOCATION, menu_type=DEFAULT_M
             'platePhotoUrl': 'https://cdn.filestackcontent.com/2X5ivrEYQvuEh30DyYot',
             'totalCount': 923,
             'totalCountUnit': 'each',
-            'primaryRecipeComponents': mock_primaryComponents,
+            'primaryRecipeComponents': [],
```
smunro-thistle, line 26: Why are these removed from the test data?

thunguy: Ah, I remember why now--there was weirdness happening with the MOCK_RECIPE_TREE_COMPONENTS data not being formatted correctly at first, so I used [] to debug the issue. Issue was that MOCK_RECIPE_TREE_COMPONENTS is getting mutated each time it’s passed back through RecipeItem to be formatted for the next mock menu item, so I just fixed that issue by adding them in as deepcopies. I’ve added two of the previous tests back in and all tests should be passing now.

smunro-thistle (approved):
Approved!

PR #84, opened 2024-09-03
(ENG-3034) Update formatted menu query to retrieve volume count

thunguy:
## Description
Formatted menu query now returns `volume` for each menu item in a menu cycle.

## Test Plan
```python
python
```

```python
from pprint import pprint

pprint(get_formatted_menu_data(dates=["2024-09-09", "2024-09-12"]))
```

- [ ] Confirm a volume count is returned for each menu item and matches the exact value in Galley

## Versioning
1.3.0 → 1.4.0

### DIFF
--- a/galley/formatted_queries.py
+++ b/galley/formatted_queries.py
@@ -519,6 +519,7 @@ def get_formatted_menu_data(
                 'recipeMenuPhotoUrl': formatted_recipe_dict.get('menuPhotoUrl', ''),
                 'recipeName': formatted_recipe_dict.get('externalName', ''),
                 'recipeProteinType': formatted_recipe_dict.get('proteinType', ''),
+                'volume': menu_item.get('volume'),
             })
         formatted_menus.append(formatted_menu)
     return formatted_menus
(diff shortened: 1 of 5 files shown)


galley/formatted_queries.py
```diff
@@ -519,6 +519,7 @@ def get_formatted_menu_data(
                 'recipeMenuPhotoUrl': formatted_recipe_dict.get('menuPhotoUrl', ''),
                 'recipeName': formatted_recipe_dict.get('externalName', ''),
                 'recipeProteinType': formatted_recipe_dict.get('proteinType', ''),
+                'volume': menu_item.get('volume'),
```
patrick-thistle, line 522: Just want to confirm that the desire here is for volume to return None if not present and we don't want to default it to 0 at this point?

patrick-thistle (approved):

galley/formatted_queries.py
thunguy: Correct. In Galley, null is the default while 0 is an intentional value.

thunguy: This is intended to send data exactly as it exists in Galley. However, if we need to store this value differently in thistle-web, we can handle that conversion at the translation layer.

PR #88, opened 2025-03-31
Update ops query to include a flattened list of all subrecipe ingredients

thunguy:
## Description
This sets up the data that OKRA needs to be able to identify allergen ingredients within the subcomponents of the primary components.

## Test Plan
In the python interactive shell, run:
```python
from galley.formatted_ops_queries import *
import json 

BASE_CODES = ['s', 'b', 'lv', 'lm', 'dv', 'dm', 't'] 
BASE_MEALS = {f'{b}{n+1}' for b in BASE_CODES for n in range(6)}
JAR_SALADS = {'ssa', 'ssb', 'ssc', 'ssd'}
SIDE_SOUPS = {'scw', 'sp',  'sm',  'sch'}
THISTLE_CLASSICS = {'cwsv', 'cwsm', 'cptv', 'cptm'}
PRODUCT_CODES_ALLOWED = BASE_MEALS | JAR_SALADS | SIDE_SOUPS | THISTLE_CLASSICS

print(json.dumps(get_formatted_ops_menu_data(dates=["2025-04-10"], product_codes_filter=PRODUCT_CODES_ALLOWED), indent=4))
```
- [x] Confirm all primary components and primary components'`recipeComponents` of type "recipe" return a flattened list of all their ingredients by the internal name e.g. "nuts, pistachio", "almond meal"
- [x] Confirm `externalName` is returned with some components' list of ingredients e.g. 24 Carrot Gold Muffin

## Versioning
1.4.0 → 1.5.0

### DIFF
--- a/galley/formatted_ops_queries.py
+++ b/galley/formatted_ops_queries.py
@@ -168,6 +168,11 @@ class RecipeItem:
             'unit': unit.get('name')
         }
 
+    def format_ingredients(self) -> List[Dict]:
+        return [
+            ingredient.get('name') for item in self.all_ingredients if (ingredient := item.get('ingredient', {}))
+        ]
+
     def to_primary_component_dict(self):
         component: Dict = {
             'allergens': self.format_allergens(),
@@ -196,14 +201,12 @@ class RecipeItem:
                     for component in self.components
                     if component.get('recipeItem')
                 ],
-                'ingredients': [
-                    ingredient.get('name') for item in self.all_ingredients if (ingredient := item.get('ingredient', {}))
-                ]
+                'ingredients': self.format_ingredients(),
             }
         return component
 
     def to_subcomponent_dict(self):
-        subcomponent: Dict = {
+        subcomponent = {
             'allergens': self.format_allergens(),
             'id': self.data.get('id'),
             # Contains the ingredients list for pre-fab ingredients
@@ -213,9 +216,7 @@ class RecipeItem:
             'usage': self.format_usage(),
         }
         if self.type == SUBRECIPE:
-            subcomponent['ingredients'] = [
-                ingredient.get('name') for item in self.all_ingredients if (ingredient := item.get('ingredient', {}))
-            ]
+            subcomponent['ingredients'] = self.format_ingredients()
         return subcomponent
 
 
--- a/galley/formatted_ops_queries.py
+++ b/galley/formatted_ops_queries.py
@@ -173,9 +173,9 @@ class RecipeItem:
             'allergens': self.format_allergens(),
             'binWeight': self.format_bin_weight(),
             'cuppingContainer': self.get_cupping_container(),
-            # External name contains the ingredients list for pre-fab ingredients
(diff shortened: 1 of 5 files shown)


galley/formatted_ops_queries.py
```diff
@@ -188,22 +191,32 @@ def to_primary_component_dict(self):
...
                         component.get('components'),
                         component.get('recipeItem').get('quantity'),
-                        component.get('recipeItem').get('unit')
+                        component.get('recipeItem').get('unit'),
                     ).to_subcomponent_dict()
                     for component in self.components
                     if component.get('recipeItem')
+                ],
+                'ingredients': [
+                    ingredient.get('name') for item in self.all_ingredients if (ingredient := item.get('ingredient', {}))
```
ErikBrown, line 200: Can `item['ingredient']` ever be present but set to null? If so, we'd need an additional check for `None` here (something like an `or {}` at the end) otherwise the `ingredient.get` would throw an error

galley/formatted_ops_queries.py
```diff
@@ -188,22 +191,32 @@ def to_primary_component_dict(self):
...
                 ]
             }
         return component
 
     def to_subcomponent_dict(self):
-        return {
+        subcomponent: Dict = {
```
ErikBrown, line 206: NB: Do we need to type this as a dict or is it inferred?

galley/formatted_ops_queries.py
```diff
@@ -188,22 +191,32 @@ def to_primary_component_dict(self):
...
+            # Contains the ingredients list for pre-fab ingredients
+            'label': self.data.get('externalName'),
             'name': self.format_name(),
             'type': self.type,
-            'usage': self.format_usage()
+            'usage': self.format_usage(),
         }
+        if self.type == SUBRECIPE:
+            subcomponent['ingredients'] = [
+                ingredient.get('name') for item in self.all_ingredients if (ingredient := item.get('ingredient', {}))
```
ErikBrown, line 217: NB: Seems like a good place to use our duplicated tag, since we're doing the same comprehension on line 200, or we can move this comprehension to its own method

ErikBrown (approved):

galley/formatted_ops_queries.py
thunguy: In that case, this resolves to:

```python
[ingredient.get('name') for item in all_ingredients if (ingredient := None)]
```

Since ingredient is `None`, the item would be filtered out. 


```python
>>> all_ingredients = [{"ingredient": None}]
>>> [ingredient.get('name') for item in all_ingredients if (ingredient := item.get('ingredient', {}))]
[]
```

galley/formatted_ops_queries.py
thunguy: Removed explicit typing.

galley/formatted_ops_queries.py
ErikBrown: Oh you're absolutely right! The `if :=` makes all those checks so elegant but trip me up sometimes

PR #92, opened 2025-08-12
Upgrade to Python 3.9 and upgrade dependencies

chase-thistle:
## Description

This is in service of shipping [thistle-web/7708](https://github.com/Thistle/thistle-web/pull/7708), a bulk dependency upgrade. The initial draft of that dependabot PR fails because the newer `segment-analytics-python` library depends on a newer version of `backoff`, and thistle-web is locked to `backoff-1.11.1` because that's what's specified in galley-sdk. This PR updates the backoff version in requirements.txt but also relaxes the backoff requirement in setup.py, allowing anything >= 2.0.

While we're at it:

* Upgrade to Python 3.9 as the minimum supported Python version for this library, given that we were already using some Python 3.9 syntax and the tests didn't pass on Python 3.8 because of it. The README had a note that Python >3.8 wouldn't work; that seems to be related to our use of a very old version of mypy that didn't support Python 3.9. So...
* Also upgrade `mypy` and `python-dotenv`
* Don't upgrade `sgqlc` (from 14.0 to 16.5) because there's a breaking change in 15.0 and we don't need to address it right now to unblock the thistle-web upgrades.

## Test Plan

- [ ] Run tests as directed in README. All tests pass.

I think that's enough testing for now. The only change that could affect application code is the `backoff` upgrade, and the test suite here includes tests of the retry behavior that still pass. We can verify that the galley integration still works when we review the PR on the thistle-web side.

## Versioning

- [ ] Version in setup.py updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@ This is a Python SDK for interacting with the graphql API published by Galley (r
 ## Installation
 To set up your environment and install the required dependencies for local development, you will may use both Python and a virtual environment tool like [virtualenv](https://virtualenv.pypa.io/en/latest/#) or [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) or the supplied [Development Container](https://containers.dev/).
 
-> :warning: Python version >3.8 is currently unsupported. You will have to downgrade Python (not recommended) or use pyenv-virtualenv to use galley-sdk.
-
 ### virtualenv steps
 ```
 $ git clone git@github.com:Thistle/galley-sdk.git
@@ -26,13 +24,13 @@ To use galley-sdk in pyenv-virtualenv:
 ```
 $ git clone git@github.com:Thistle/galley-sdk.git
 $ cd galley-sdk
-$ pyenv install 3.8.13
-$ pyenv virtualenv 3.8.13 py38
-$ pyenv activate py38
+$ pyenv install 3.9
+$ pyenv virtualenv 3.9 py39
+$ pyenv activate py39
 $ pip install -r requirements.txt
 ```
 
-> Note: When using galley-sdk in the future, make sure you run `pyenv activate py38` beforehand to be in the correct virtual envrionment.
+> Note: When using galley-sdk in the future, make sure you run `pyenv activate py39` beforehand to be in the correct virtual envrionment.
 
 ### Using the supplied Development Container
 To use the development container you will need a docker environment installed.
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,8 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/python-3/.devcontainer/base.Dockerfile
 
-# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+# [Choice] Python version (use -bookworm or -bullseye variants on local arm64/Apple Silicon): 3, 3.13, 3.12, 3.11, 3.10, 3.9, 3-bookworm, 3.13-bookworm, 3.12-bookworm, 3.11-bookworm, 3.10-bookworm, 3.9-bookworm, 3-bullseye, 3.13-bullseye, 3.12-bullseye, 3.11-bullseye, 3.10-bullseye, 3.9-bullseye, 3-buster, 3.12-buster, 3.11-buster, 3.10-buster, 3.9-buster
 ARG VARIANT="3.10-bullseye"
-FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+FROM mcr.microsoft.com/devcontainers/python:1-${VARIANT}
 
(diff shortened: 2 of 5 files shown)


galley/scripts/bulk_add_preparation_to_ingredient_usages.py
```diff
@@ -65,10 +65,11 @@ def add_preparations_to_recipe_items_from_ingredient_csv(filename: str, dry_run=
     items_by_prep = defaultdict(list)
 
     for item in recipe_items:
-        if item.get('preparation') in PreparationEnum._member_names_:
-            prep_id = PreparationEnum[item.get('preparation')].value
```
chase-thistle, line 69: > galley/scripts/bulk_add_preparation_to_ingredient_usages.py:69: error: Enum index should be a string (actual index type "Union[str, Any, None]")  [misc]

galley/scripts/bulk_add_preparation_to_ingredient_usages.py
```diff
@@ -80,11 +81,12 @@ def add_preparations_to_recipe_items_from_recipe_csv(filename: str, dry_run=True
...
             recipes.append(row)
 
-    items_by_prep = defaultdict(list)
+    items_by_prep: dict = defaultdict(list)
 
     for recipe in recipes:
-        if recipe.get('preparation') in PreparationEnum._member_names_:
-            prep_id = PreparationEnum[recipe.get('preparation')].value
-            if prep_id:
-                items_by_prep[prep_id].extend(recipe.get('recipe_item_ids').split(','))
```
chase-thistle, line 89: > galley/scripts/bulk_add_preparation_to_ingredient_usages.py:83: error: Need type annotation for "items_by_prep"  [var-annotated]
>
> galley/scripts/bulk_add_preparation_to_ingredient_usages.py:87: error: Enum index should be a string (actual index type "Union[str, Any, None]")  [misc]
>
> galley/scripts/bulk_add_preparation_to_ingredient_usages.py:89: error: Item "None" of "Union[str, Any, None]" has no attribute "split"  [union-attr]

Easiest fix was to do string casts, and I also switched to a try-except approach to the enum lookup instead of relying on `_member_names_`, which is a private Enum API.

galley/scripts/custom_preparation_utils.py
```diff
@@ -204,17 +207,17 @@ def delete_ingredient_usage_preparations_by_preparation_ids(
             for recipe_item_preparation in recipe_item_preparations
             if (
                 ingredient := recipe_item_preparation.get("recipeItem", {}).get("ingredient")
-                and ingredient["id"] in include_ingredient_ids
```
chase-thistle, line 207: > galley/scripts/custom_preparation_utils.py:207: error: Cannot determine type of "ingredient"  [has-type]
>
> galley/scripts/custom_preparation_utils.py:207: error: Value of type "Union[Any, bool]" is not indexable  [index]
>
> galley/scripts/custom_preparation_utils.py:207: error: Name "ingredient" is used before definition  [used-before-def]
>
> galley/scripts/custom_preparation_utils.py:216: error: Value of type "Union[Any, bool]" is not indexable  [index]
>
> galley/scripts/custom_preparation_utils.py:216: error: Name "ingredient" is used before definition  [used-before-def]

Subtle difference here! Before, Python could have interpreted this as:

```python
ingredient := (
  recipe_item_preparation.get("recipeItem", {}).get("ingredient")
  and ingredient["id"] in include_ingredient_ids
)
```

Now it's unambiguous; now only the first line is part of the assignment, and `ingredient["id"] in include_ingredient_ids` gets evaluated after. This makes mypy happy.

My guess is that Python _didn't_ actually interpret it that way — that in fact it knew it couldn't evaluate `ingredient["id"]` until the assignment was done, so it did the assignment first.

Or maybe it did, and this is totally broken, and it just hasn't come up because this is old script code that never gets run? I don't know.

tests/test_ops_formatted_queries.py
```diff
@@ -202,29 +202,29 @@ def test_format_recipe_allergen_data_empty(self):
...
+            "ingredient": None,
+            "subRecipe": {
+                "dietaryFlagsWithUsages": None
+            }
         }
         result = RecipeItem(recipeitem=recipeitem).format_allergens()
         self.assertEqual(result, [])
 
-    def test_format_recipe_allergen_data_None(self):
+    def test_format_ingredient_allergen_data_empty(self):
```
chase-thistle, line 216: `test_format_recipe_allergen_data_None` was a duplicate test name but it looks like we actually wanted to have two tests here, one for the `None` case for recipes and one for the `None` case for ingredients. Fortunately they both pass already.

smunro-thistle (approved):
Nice! Thanks for the explanations of errors/fixes - that gave me a lot of confidence in reviewing the changes. Just the one comment about the missing step for using virtualenv instead of pyenv in the README, but I think that is NB since that was an issue previously, too.

README.md
```diff
@@ -6,8 +6,6 @@ This is a Python SDK for interacting with the graphql API published by Galley (r
 ## Installation
 To set up your environment and install the required dependencies for local development, you will may use both Python and a virtual environment tool like [virtualenv](https://virtualenv.pypa.io/en/latest/#) or [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) or the supplied [Development Container](https://containers.dev/).
 
-> :warning: Python version >3.8 is currently unsupported. You will have to downgrade Python (not recommended) or use pyenv-virtualenv to use galley-sdk.
-
 ### virtualenv steps
```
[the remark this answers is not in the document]
chase-thistle: Good suggestion! Added a line here.

